PR #10, opened 2024-03-15
Remove put_objects line so datasets create correctly

douglas1850:
*Issue #, if available:* N/A

*Description of changes:* Datasets should be not be created using the put_object function; this is resulting in the deletion of dataset "files" deleting the entire dataset itself.

### DIFF
--- a/backend/src/ml_space_lambda/dataset/lambda_functions.py
+++ b/backend/src/ml_space_lambda/dataset/lambda_functions.py
@@ -197,7 +197,6 @@ def create_dataset(event, context):
         raise Exception("Dataset headers do not match expected type and scope.")
 
     if not dataset_dao.get(scope, dataset_name):
-        s3.put_object(Bucket=env_variables["DATA_BUCKET"], Key=directory_name)
         dataset_location = f's3://{env_variables["DATA_BUCKET"]}/{directory_name}'
         dataset = DatasetModel(
             scope=scope,
--- a/backend/test/dataset/test_create_dataset.py
+++ b/backend/test/dataset/test_create_dataset.py
@@ -18,7 +18,6 @@ import json
 from unittest import mock
 
 import pytest
-from botocore.exceptions import ClientError
 
 from ml_space_lambda.data_access_objects.dataset import DatasetModel
 from ml_space_lambda.enums import DatasetType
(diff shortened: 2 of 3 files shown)


backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -197,7 +197,6 @@ def create_dataset(event, context):
         raise Exception("Dataset headers do not match expected type and scope.")
 
     if not dataset_dao.get(scope, dataset_name):
-        s3.put_object(Bucket=env_variables["DATA_BUCKET"], Key=directory_name)
```
dustins, line 200: Was this always broken? Why does this cause it to fail, does it just create an exception?

douglas1850:
Looks like I need to update the unit test too to get this to pass the checks, working on it.

backend/src/ml_space_lambda/dataset/lambda_functions.py
douglas1850: This was putting an actual object at the location of the dataset path in the create function. Removing this fixes deleting files from datasets without breaking the creation of datasets.

I confirmed this by running the suite of dataset integtests and they all pass.

douglas1850:
Now it looks like the linter is failing for the unit test file `test_create_dataset`. It's passing with the bb format linter, so I'm not sure why it's failing here...

backend/src/ml_space_lambda/dataset/lambda_functions.py
KyleLilly: This was relatively new. It was added with support for empty datasets but explicitly putting an object with a key of the "directory" resulted in an empty object existing in s3 that matched the dataset path. Normally everything in s3 has a key and there is no actual directories. For the sake of visualization `s3://bucket/my/nested/file` will be rendered as two "directories" `my` and `nested` but there is only one object in the bucket `my/nested/file`. This created an empty object with the key `my/nested/` and then when you added a file we wind up with 2 objects in the bucket `my/nested/` and `my/nested/file`. Having this extra object was the cause of the bug.

dustins: I see so it wasn't breaking it, it just created an empty file. Thanks!

dustins (approved):

backend/test/dataset/test_create_dataset.py
```diff
@@ -114,37 +107,9 @@ def test_create_dataset_already_exists(mock_s3, mock_dataset_dao):
     expected_response = generate_html_response(400, error_message)
 
     assert lambda_handler(mock_event, mock_context) == expected_response
-    mock_s3.put_object.assert_not_called()
 
 
-@mock.patch("ml_space_lambda.dataset.lambda_functions.dataset_dao")
-@mock.patch("ml_space_lambda.dataset.lambda_functions.s3")
-def test_create_dataset_client_error(mock_s3, mock_dataset_dao):
```
KyleLilly, line 122: We _could_ still get throttle from our dynamo during our DAO call but in this particular case I think we have sufficient coverage.

KyleLilly (approved):

PR #12, opened 2024-03-15
[AIML-ADC-7385] Adding devcontainer and pre-commit

AlejandroRigau:
*Issue #, if available:* AIML-ADC-7385

*Description of changes:*
- Adding a devcontainer to facilitate development and ensure all developer have the same environment
- Adding pre-commit hooks to facilitate clean and maintainable code


pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[tool.black]
+line-length = 130
+
+[tool.flake8]
+max-line-length = 130
+
+[tool.isort]
+profile = "black"
+line_length = 130
```
KyleLilly, line 9: These line lengths don't match the values in `setup.cfg` or the existing action. It's also not configured to ignore the warnings we're ignoring. Do we need this file?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
+repos:
+
+- repo: https://github.com/psf/black
+  rev: 23.1.0
```
KyleLilly, line 4: Similar comment about matching what's in our action. If we're pinning a version here we likely need to pin the same version there.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
KyleLilly, line 1: General ignorance - is this going to conflict with the existing org level dependabot configuration?

AlejandroRigau: This file was autogenerated for me. Ill investigate

pyproject.toml
AlejandroRigau: pyproject.toml seems to fulfill the same purpose of the `setup.cfg` so i will look for a way to use that configuration file instead.

.pre-commit-config.yaml
AlejandroRigau: ill verify this

AlejandroRigau:
I removed the dependabot configuration so that it does not interfere with the org level dependabot. 
Updated  pre-commit hooks to match github actions and setup.cfg.

Sending for approval again

KyleLilly (approved):

.devcontainer/post_create_command.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+npm install
+cd frontend/ && npm install
+cd ..
```
dustins, line 5: Just a note I like what we did for our [lambda_dependencies](https://github.com/awslabs/mlspace/blob/main/lambda_dependencies/jwt/create.sh#L6) where we use `pushd new/path` and then `popd` to get back to the original directory.  It doesn't matter as much here because we're only going one directory deep so we're not doing something like `../../..` but sharing for future knowledge where it makes more sense.

dustins (approved):

PR #15, opened 2024-03-19
Empty dataset support

douglas1850:
*Issue #, if available:*

*Description of changes:* Changes to allow for datasets to be created at the time of translate/transform/training jobs being started. Also made a few small fixes to clean up things the linter was picking up like undefined values, deleting non-optional values, etc.

### DIFF
--- a/frontend/src/entities/dataset/dataset.service.ts
+++ b/frontend/src/entities/dataset/dataset.service.ts
@@ -1,9 +1,12 @@
 /**
   Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
   Licensed under the Apache License, Version 2.0 (the "License").
   You may not use this file except in compliance with the License.
   You may obtain a copy of the License at
+
       http://www.apache.org/licenses/LICENSE-2.0
+
   Unless required by applicable law or agreed to in writing, software
   distributed under the License is distributed on an "AS IS" BASIS,
   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/frontend/src/entities/dataset/dataset.service.ts
+++ b/frontend/src/entities/dataset/dataset.service.ts
@@ -1,12 +1,9 @@
 /**
   Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-
   Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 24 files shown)


AlejandroRigau (dismissed):

frontend/src/entities/dataset/dataset.service.ts
```diff
@@ -185,28 +185,37 @@ export async function uploadFiles (
...
+    let unexpectedError = false;
+    createDataset(dataset).catch((error) => {
+        const expectedError = `Bad Request: Dataset ${dataset.name} already exists.`;
+        // Any error that the dataset already existing is unexpected
+        unexpectedError = expectedError !== error.response.data;
+    });
+    
+    return unexpectedError;
+};
+
```
KyleLilly, line 218: Why not just swallow expected errors and throw the unexpected ones? Is there a reason we're returning it and relying on the caller to know that/do something with it?

frontend/src/entities/jobs/hpo/create/hpo-job-create.tsx
```diff
@@ -351,7 +354,29 @@ export function HPOJobCreate () {
...
+                    if (split[3] === 'private') {
+                        datasetType = DatasetType.PRIVATE;
+                    }
+                    const datasetName = split[6].replace('"', '');
+                    const newDataset = {
+                        name: datasetName,
+                        description: `Dataset created as part of the HPO job: ${state.form.HyperParameterTuningJobName}`,
+                        type: datasetType,
+                        scope: scope
+                    } as IDataset;
```
KyleLilly, line 370: Cab we use `datasetFromS3Uri()` here?

frontend/src/entities/dataset/dataset.service.ts
douglas1850: You're thinking I do the expected error check and if it's an unexpected error I just re-raise it, and then have the call to `createDatasetHandleAlreadyExists` be invoked in a try/catch right?

frontend/src/entities/jobs/hpo/create/hpo-job-create.tsx
douglas1850: Good call, wasn't aware of this function but it serves the same purpose.

KyleLilly (approved):

PR #17, opened 2024-03-21
Clarifies Setup Instructions

LightSeekerSC:
*Description of changes:* Updates README instructions to clarify for brand new developers


README.md
```diff
@@ -70,6 +70,8 @@ Update the values in `lib/constants.ts` based on your specific deployment needs.
 ### Creating a production optimized web app build
 In addition to updating the necessary parameters in the CDK constants file you will also need to create a production build of the web application. You can do this by changing to the web application directory (`frontend/`) and running:
 ```
+# From /frontend directory
+cd frontend
```
dustins, line 74: I get wanting this clarification but doesn't this sound like you're in `/frontend` and then trying to `cd` into it?

README.md
```diff
@@ -83,9 +85,11 @@ You will also need to set `OIDC_URL` and `OIDC_CLIENT_NAME` with the correct val
 
 The MLSpace application is a standard CDK application and can be deployed just as any CDK application is deployed:
 ```
+# From / root directory
```
dustins, line 88: I think this should say "project root" or something because `/` and `root` invoke the root of the file system.

README.md
KyleLilly: Yeah I think the wording above:
> You can do this by changing to the web application directory (`frontend/`) and running:

implies the user should be in the directory when running this command but here we're saying from the `/frontend` run `cd frontend` which would fail.

AlejandroRigau:
lgtm after previous comments have been addressed

README.md
LightSeekerSC: Yeah that is fair. That does make sense and I understand why the comment and command are conflicting. I think that having the cd in the codeblock helps to bring attention to what directory you should be in. My personal flaw is that I scan for codeblocks and don't always read text that would prompt me for what directory to be in, which is why I added this as I feel like it clarifies for more people without detracting. It makes it easier for anyone to just copy and paste the code without having to back-track read for context on what directory to be in. Generally I assume all commands should be run from the project root unless there is a cd command, but maybe that is just me.

dustins: I think that is fair but then the wordings needs to be changed to something like this.

> ### Creating a production optimized web app build
> In addition to updating the necessary parameters in the CDK constants file you will also need to create a production build of the web application. You can do this from the web application directory `frontend/`.
> ```
> # Build production web application
> cd frontend
> npm run clean && npm install
> ```

LightSeekerSC:
Accidentally uploaded constants.ts for Rev2. Removed constants and added project root comment changes

AlejandroRigau (approved):

KyleLilly (approved):

PR #19, opened 2024-03-22
Migrate emr to resource metadata

douglas1850:
*Issue #, if available:*

*Description of changes:*


AlejandroRigau:
I meant to request changes rather than click on comment. Some of my above comments are blocking.

backend/src/ml_space_lambda/emr/lambda_functions.py
```diff
@@ -188,13 +195,14 @@ def create(event, context):
 
         clusters = list_clusters_for_project(
             emr=emr,
-            prefix=project_name,
+            prefix=None,
```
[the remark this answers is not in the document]
douglas1850: Good call, removed it

backend/src/ml_space_lambda/report/lambda_functions.py
```diff
@@ -208,31 +208,30 @@ def _get_training_jobs(resource_get_func: Callable[..., PagedMetadataResults], a
...
 def _get_emr_clusters(
-    cluster_prefix: Optional[str] = None,
-    cluster_owner: Optional[str] = None,
+    resource_get_func: Callable[..., PagedMetadataResults],
+    args: Dict[str, any],
     termination_records: Optional[List[ResourceSchedulerModel]] = None,
 ) -> List[Dict[str, Any]]:
-    response = list_clusters_for_project(emr, prefix=cluster_prefix, fetch_all=True)
-
+    logger.info("-----Inside the EMR function-----")
```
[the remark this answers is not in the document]
douglas1850: we do not, removing log statements

lib/constants.ts
```diff
@@ -33,12 +33,12 @@ export const NOTEBOOK_PARAMETERS_FILE_NAME = 'notebook-params.json';
 export const PERMISSIONS_BOUNDARY_POLICY_NAME = '';
 
 // This could be something like Admin or a dedicated role for KMS Key Management
-export const KEY_MANAGER_ROLE_NAME = '';
+export const KEY_MANAGER_ROLE_NAME = 'Admin';
 
 // Account ID is appended to s3 buckets to ensure uniqueness within region
-export const AWS_ACCOUNT = '';
+export const AWS_ACCOUNT = '992382812452';
```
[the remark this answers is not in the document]
douglas1850: Good catch, removing

backend/src/ml_space_lambda/project/lambda_functions.py
```diff
@@ -230,14 +230,11 @@ def remove_user(event, context):
 
     if Permission.PROJECT_OWNER not in project_member.permissions or total_project_owners(project_user_dao, project_name) > 1:
         # Terminate any running EMR Clusters the user owns that are associated with the project
+        clusters = resource_metadata_dao.get_all_for_project_by_type(project_name, ResourceType.EMR_CLUSTER, fetch_all=True)
         cluster_ids = []
```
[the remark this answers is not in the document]
douglas1850: Good suggestion! It looks like for list comprehension the conditional goes on the end if you don't want to have an `else`

```
cluster_ids = [cluster.id for cluster in clusters.records if cluster.user == username]
```

backend/src/ml_space_lambda/migration/lambda_functions.py
```diff
@@ -308,3 +312,39 @@ def _sync_hpo_jobs(env_variables):
...
+                        try:
+                            resource_metadata_dao.upsert_record(
+                                cluster["Id"],
+                                ResourceType.EMR_CLUSTER,
+                                owner,
+                                project,
+                                metadata,
+                            )
+                        except Exception:
+                            logger.warn(f'Error generating resource metadata for Model: {cluster["Name"]}')
```
KyleLilly, line 350:
```suggestion
                            logger.warn(f'Error generating resource metadata for EMR Cluster: {cluster["Name"]}')
```

backend/src/ml_space_lambda/report/lambda_functions.py
```diff
@@ -28,7 +28,7 @@
 from ml_space_lambda.data_access_objects.resource_scheduler import ResourceSchedulerDAO, ResourceSchedulerModel
 from ml_space_lambda.data_access_objects.user import UserDAO, UserModel
 from ml_space_lambda.enums import Permission, ResourceType
-from ml_space_lambda.utils.common_functions import api_wrapper, get_tags_for_resource, list_clusters_for_project, retry_config
+from ml_space_lambda.utils.common_functions import api_wrapper, retry_config
```
KyleLilly, line 31: Not totally related to this PR but can you remove the EMR client (line 42)? It shouldn't be used any more. The sagemaker client (line 39) can also be removed.

backend/src/ml_space_lambda/resource_metadata/lambda_functions.py
```diff
@@ -306,6 +309,40 @@ def _process_endpoint_config_event(details):
...
+        resource_metadata_dao.delete(cluster_id, ResourceType.EMR_CLUSTER)
+    else:
+        cluster_details = emr.describe_cluster(ClusterId=cluster_id)
+        tags = _parse_tags(cluster_details["Cluster"]["Tags"])
+        if "system" in tags and tags["system"] == env_variables["SYSTEM_TAG"]:
+            if "user" in tags and "project" in tags:
+                user = tags["user"]
+                project = tags["project"]
+                metadata = {
+                    "CreationTime": cluster_details["Cluster"]["Status"]["Timeline"]["CreationDateTime"],
```
KyleLilly, line 327: Just want to double check we don't need to call `_convert_timestamp()` on these. I know the APIs aren't all consistent in how they return data so this may be fine.

backend/src/ml_space_lambda/utils/common_functions.py
```diff
@@ -149,7 +149,6 @@ def get_tags_for_resource(sagemaker, arn: str):
 
 def list_clusters_for_project(
```
KyleLilly, line 150: Is there a reason this wasn't switched over to use resource metadata and is still relying on the EMR client?

KyleLilly (changes requested):
It looks like this is missing updates to the authorizer logic as well. We should be able to remove the EMR client there and just rely on the resource metadata for determining authZ for the related actions.

backend/src/ml_space_lambda/utils/common_functions.py
douglas1850: The only place we're using it at this point is when we create the EMR cluster. We grab the list of clusters created in the last 20 seconds then we create an entry in the resource scheduler for each one of them.

```
        clusters = list_clusters_for_project(
            emr=emr,
            created_after=datetime.datetime.fromtimestamp(time.time() - 20),
        )
        for cluster in clusters["records"]:
            if cluster["Name"] == cluster_name:
                resource_scheduler_dao.create(
                    ResourceSchedulerModel(
                        resource_id=cluster["Id"],
                        resource_type=ResourceType.EMR_CLUSTER,
                        termination_time=termination_time,
                        project=project_name,
                    )
                )
                break
```

I'm not positive that this EMR cluster would exist in the resource metadata table yet since we'd have a run between the `run_job_flow` execution triggering the CW event logic, versus hitting this line.

backend/src/ml_space_lambda/report/lambda_functions.py
douglas1850: Good call, neither is in use.

backend/src/ml_space_lambda/resource_metadata/lambda_functions.py
douglas1850: >Creation time
Mar 20, 2024, 16:47:19 EDT

This is what the date looks like in the console right now, so it seems alright

backend/src/ml_space_lambda/utils/common_functions.py
KyleLilly: Ok I forgot about that. `run_job_flow` returns a cluster arn but not a cluster id and we can't describe a cluster without the id so that's what we use for termination. Seems like we need to just leave this for now.

backend/src/ml_space_lambda/utils/common_functions.py
```diff
@@ -150,7 +150,6 @@ def get_tags_for_resource(sagemaker, arn: str):
...
```
KyleLilly, line 150: Name is probably a bit misleading at this point and it could be confusing given that  it's in common utils but in actuality has a very specific usecase.

KyleLilly (dismissed):

KyleLilly (approved):

backend/test/authorizer/test_authorizer.py
```diff
@@ -2240,10 +2213,10 @@ def test_manage_project_sagemaker_resource_boto_error(
         "ResponseMetadata": {"HTTPStatusCode": 400},
     }
     mock_project_dao.get.return_value = MOCK_PROJECT
-    if path_param_key in ["notebookName", "endpointName", "modelName", "endpointConfigName"]:
+    if path_param_key in ["notebookName", "endpointName", "modelName", "endpointConfigName", "clusterId"]:
         mock_resource_metadata_dao.get.side_effect = ClientError(error_msg, "GetItem")
-    if path_param_key == "clusterId":
-        mock_emr.describe_cluster.side_effect = ClientError(error_msg, "DescribeCluster")
+    # if path_param_key == "clusterId":
+    #     mock_emr.describe_cluster.side_effect = ClientError(error_msg, "DescribeCluster")
```
dustins, line 2219: Do we need to keep these commented lines?

AlejandroRigau:
LGTM after you address Dustins comment and merge from main.

backend/test/authorizer/test_authorizer.py
douglas1850: ah yeah I missed that during testing, will remove.

KyleLilly (approved):

dustins (approved):

PR #20, opened 2024-03-25
Adds Semantic Segmentation Algorithm And Consolidates Validation

LightSeekerSC:
*Issue #, if available:* https://issues.amazon.com/issues/AIML-ADC-7257

*Description of changes:*
- Adds semantic segmentation algorithm for use in MLSpace
- Consolidates other numeric validators
  - Some numeric validators were left as wrappers (can remove if desired)

Recording 2024-03-25 at 14 55 15

### DIFF
--- a/frontend/src/entities/jobs/algorithms.tsx
+++ b/frontend/src/entities/jobs/algorithms.tsx
@@ -24,9 +24,11 @@ import deepARMetadata from '../../shared/algorithmMetadata/forecastingDeepAR.jso
 import object2Vec from '../../shared/algorithmMetadata/object2Vec.json';
 import imageClassification from '../../shared/algorithmMetadata/imageClassification.json';
 import factorizationMachines from '../../shared/algorithmMetadata/factorizationMachines.json';
+import semanticSegmentationMetadata from '../../shared/algorithmMetadata/semanticSegmentation.json';
 import React, { ReactNode } from 'react';
-import { floatIntervalValidator, integerIntervalValidator, IntervalStrategy } from '../../shared/validation/helpers/numbers';
 import xgboostMetadata from '../../shared/algorithmMetadata/xgboost.json';
+
+const autoContext = z.literal('auto');
 
 export type MetricDefinition = {
     metricName: string;
@@ -52,10 +54,10 @@ export type Algorithm = {
 };
 
 export enum HyperparameterType {
-    CATEGORICAL = 'categorical',
-    CONTINUOUS = 'continuous',
-    INTEGER = 'integer',
-    STATIC = 'static',
+    CATEGORICAL = 'Categorical',
+    CONTINUOUS = 'Continuous',
+    INTEGER = 'Integer',
+    STATIC = 'Static',
 }
 
 export type Hyperparameter = {
@@ -71,83 +73,114 @@ export type Hyperparameter = {
     converter?: (input: any) => any;
 };
 
-const floatValidator = (required = false, positive = false) => {
-    let number_context = z.number({
-        invalid_type_error: `Must be a float${required ? ' and is required.' : ''}`,
+const floatValidator = (fieldName: string, required = false, positive = false) => {
+    return numberValidator(fieldName, {
+        isFloat: true,
+        required: required,
+        min: positive ? 0 : Number.NEGATIVE_INFINITY,
+        includeMin: false // If positive, can't be zero, will never reach Number.NEGATIVE_INFINITY regardless
     });
-
-    if (positive) {
-        number_context = number_context.min(0, { message: 'Must be at least 0' });
-    }
-
-    let context: any = z.preprocess(Number, number_context);
-    if (!required) {
-        context = context.optional().or(z.literal(''));
-    }
-    return context;
 };
 
-const positiveIntMessage = (field: string, required = true) =>
-    `${field} ${required ? 'is required and ' : ''}must be a positive integer`;
-
-const positiveIntValidator = (fieldName: string, required = true, minimum = 1) => {
-    const number_context = z.number().min(minimum, {
-        message: `${positiveIntMessage(fieldName, required)} of at least ${minimum}`,
-    });
-
-    let context: any = z.preprocess(Number, number_context);
-    if (!required) {
-        context = context.optional().or(z.literal(''));
-    }
-    return context;
+const positiveIntValidator = (fieldName: string, required = true, minimum = 1, inputProps : rangeValidatorProps = {}) => {
+    return numberValidator(fieldName, 
+        {
+            min: minimum,
+            required: required,
+            ...inputProps
+        })
 };
 
-const rangeValidator = (
+type rangeValidatorProps = {
+    min?: number;
+    max?: number;
+    isFloat?: boolean;
+    required?: boolean;
+    allowAuto?: boolean;
+    includeMin?: boolean;
+    includeMax?: boolean;
+    alternateValues?: string[];
+}
+
+/**
+ * Creates a validator for ensuring a number meets the required conditions
+ * 
+ * @param fieldName Name of the field being validated
+ * @param inputProps Properties that ensure the validity of the field
+ * @returns A validator for the field that ensures the desired requirements
+ */
+const numberValidator = (
     fieldName: string,
-    min: number,
-    max: number,
-    isFloat = false,
-    required = true,
-    allowAuto = false
+    inputProps?: rangeValidatorProps
 ) => {
-    const validationMessage = `${fieldName} must be ${
-        isFloat ? 'a float' : 'an integer'
-    } value in the range [${min},${max}]${allowAuto ? ' or be \'auto\'' : ''}`;
+    // Create properties by merging the defaults with the provided input properties
+    const props = {
+        min: Number.NEGATIVE_INFINITY, // Specifies the smallest number allowed (default is infinitely negative to allow for any number)
+        max: Number.POSITIVE_INFINITY, // Specifies the largest number allowed (default is ifinitely positive to allow for any number)
+        isFloat: false, // The number is assumed to be an Integer unless specified (default is that it is an Integer)
+        required: false, // Specify whether this field is required for submission and must not be empty (default is that the field is not required)
+        allowAuto: false, // Allows for the text option of "auto" instead of a number (default is that "auto" is not valid)
+        includeMin: true, // Whether the specified minimum is a valid number option (default is that it is valid)
+        includeMax: true, // Whether the specified maximum is a valid number option (default is that it is valid)
+        ...inputProps
+    }
 
-    const numberContext = z.preprocess(
-        Number,
-        z.number().gte(min, { message: validationMessage }).lte(max, { message: validationMessage })
-    );
+    // Create range message
+    let rangeMessage = '';
+    if (props.min !== Number.NEGATIVE_INFINITY && props.max !== Number.POSITIVE_INFINITY){
+        // Create range message
+        rangeMessage = ` value in the range ${props.includeMin?'[':'('}${props.min},${props.max}${props.includeMax?']':')'}`
+    } else if (props.max === Number.POSITIVE_INFINITY) {
+        // If there is a min, but not a max, create greater than message
+        rangeMessage = ` value greater than ${props.includeMin?'or equal to ':''}${props.min}`
+    } else if (props.min === Number.NEGATIVE_INFINITY) {
+        // If there is a max, but not a min, create less than message
+        rangeMessage = ` value less than ${props.includeMax?'or equal to ':''}${props.max}`
+    }
 
-    const autoContext = z.literal('auto');
+    // Create validation message for errors
+    const validationMessage = `${fieldName}${props.required ? ' is required and' : '' } must be ${
+        props.isFloat ? 'a float' : 'an integer'
+    }${rangeMessage}${props.allowAuto ? ' or be \'auto\'' : ''}${props.alternateValues ? ' or be one of the following values: ' + props.alternateValues.map(value => `'${value}'`).join(', ') : ''}`;
 
+
+    // Create base number validator
+    let numberContext = z.coerce.number({
+        invalid_type_error: validationMessage,
+    });
+
+    // Add min/max check based on min and max inclusions
+    numberContext = props.includeMin ? numberContext.gte(props.min, { message: validationMessage }) : numberContext.gt(props.min, { message: validationMessage });
+    numberContext = props.includeMax ? numberContext.lte(props.max, { message: validationMessage }) : numberContext.lt(props.max, { message: validationMessage });
+
+    // Create a wrapper validator that allows for possible text values
     let wrapperContext: any = numberContext;
 
-    if (allowAuto) {
-        wrapperContext = numberContext.or(autoContext);
+    // Appends auto or other alternative non-number allowed values
+    if (props.allowAuto) {
+        wrapperContext = z.union([wrapperContext, autoContext], {errorMap: () => ({ message: validationMessage})});
+    }
+    if (props.alternateValues) {
+        props.alternateValues.forEach((value) => {
+            wrapperContext = z.union([wrapperContext, z.literal(value)], {errorMap: () => ({ message: validationMessage})});
+        })
     }
 
-    if (!required) {
+    if (!props.required) {
         wrapperContext = wrapperContext.optional();
     }
 
     return wrapperContext;
 };
 
-const positiveIntOrAutoValidator = (fieldName: string, allowIgnore = false) => {
-    const numberContext = z
-        .number()
-        .min(1, { message: positiveIntMessage(fieldName) })
-        .or(z.literal('auto'));
-    if (allowIgnore) {
-        numberContext.or(z.literal('ignore'));
-    }
-
-    const wrapperContext = z.literal('auto').or(z.preprocess(Number, numberContext));
-    if (allowIgnore) {
-        wrapperContext.or(z.literal('ignore'));
-    }
-    return wrapperContext;
+const positiveIntOrAutoValidator = (fieldName: string, alternateValues:string[] = [], allowIgnore = false) => {
+    return numberValidator(fieldName,
+        {
+            min: 1,
+            allowAuto: true,
+            required: !allowIgnore,
+            alternateValues: alternateValues
+        })
 };
 
 const linearIntProperties = {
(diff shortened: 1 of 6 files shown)


frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -24,10 +24,12 @@ import deepARMetadata from '../../shared/algorithmMetadata/forecastingDeepAR.jso
 import object2Vec from '../../shared/algorithmMetadata/object2Vec.json';
 import imageClassification from '../../shared/algorithmMetadata/imageClassification.json';
 import factorizationMachines from '../../shared/algorithmMetadata/factorizationMachines.json';
+import semanticSegmentationMetadata from '../../shared/algorithmMetadata/semanticSegmentation.json';
 import React, { ReactNode } from 'react';
-import { floatIntervalValidator, integerIntervalValidator, IntervalStrategy } from '../../shared/validation/helpers/numbers';
```
KyleLilly, line 28: Are these helpers used at all now? I thought the outcome of our discussion was that we wanted validators we could use across the application and all of our validators would exist in this helper and then be referenced from here as well as various create forms. If that's not the case let's remove that helper file completely to avoid the issue that originally prompted this change which was that we had multiple validators that appeared to offer duplicate functionality and it wasn't clear when to use what.

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -52,10 +54,10 @@ export type Algorithm = {
...
 
 export enum HyperparameterType {
-    CATEGORICAL = 'categorical',
-    CONTINUOUS = 'continuous',
-    INTEGER = 'integer',
-    STATIC = 'static',
+    CATEGORICAL = 'Categorical',
+    CONTINUOUS = 'Continuous',
+    INTEGER = 'Integer',
+    STATIC = 'Static',
```
KyleLilly, line 60: Any particular reason for this change?

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -71,83 +73,114 @@ export type Hyperparameter = {
...
-    if (positive) {
-        number_context = number_context.min(0, { message: 'Must be at least 0' });
-    }
-
-    let context: any = z.preprocess(Number, number_context);
-    if (!required) {
-        context = context.optional().or(z.literal(''));
-    }
-    return context;
+const positiveIntValidator = (fieldName: string, required = true, minimum = 1, inputProps : rangeValidatorProps = {}) => {
```
KyleLilly, line 85: We may need to update the lint/formatter rule if this wasn't picked up.
```suggestion
const positiveIntValidator = (fieldName: string, required = true, minimum = 1, inputProps: rangeValidatorProps = {}) => {
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -71,83 +73,114 @@ export type Hyperparameter = {
...
 };
 
-const positiveIntMessage = (field: string, required = true) =>
-    `${field} ${required ? 'is required and ' : ''}must be a positive integer`;
+type rangeValidatorProps = {
+    min?: number;
+    max?: number;
+    isFloat?: boolean;
+    required?: boolean;
+    allowAuto?: boolean;
```
KyleLilly, line 99: Is there a reason for `allowAuto` and `alternateValues`? Can we just include `auto` in the alternate values array?

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -71,83 +73,114 @@ export type Hyperparameter = {
...
+    }
 
-    let context: any = z.preprocess(Number, number_context);
-    if (!required) {
-        context = context.optional().or(z.literal(''));
+    // Create range message
+    let rangeMessage = '';
+    if (props.min !== Number.NEGATIVE_INFINITY && props.max !== Number.POSITIVE_INFINITY){
+        // Create range message
+        rangeMessage = ` value in the range ${props.includeMin?'[':'('}${props.min},${props.max}${props.includeMax?']':')'}`
```
KyleLilly, line 132: Similar comment to the above if the linter isn't complaining about the lack of spaces between in your ternary then we probably need to update the configuration. In the meantime just manually add the spaces.

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -71,83 +73,114 @@ export type Hyperparameter = {
...
-    if (allowIgnore) {
-        numberContext.or(z.literal('ignore'));
-    }
-
-    const wrapperContext = z.literal('auto').or(z.preprocess(Number, numberContext));
-    if (allowIgnore) {
-        wrapperContext.or(z.literal('ignore'));
-    }
-    return wrapperContext;
+const positiveIntOrAutoValidator = (fieldName: string, alternateValues:string[] = [], allowIgnore = false) => {
```
KyleLilly, line 176: `allowIgnore` wasn't related to the field being required in the original implementation. The `alternatveValues` parameter serves the same purpose of the original allow ignore. It looks like you're primarily calling it with  `alternateValues` of `['ignore','']` which means you're allowing an empty string which would mean it's optional. The original implementation didn't allow for that so I just want to double check that change is intended.
```suggestion
const positiveIntOrAutoValidator = (fieldName: string, alternateValues: string[] = [], required = false) => {
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -71,83 +73,114 @@ export type Hyperparameter = {
...
+    return numberValidator(fieldName,
+        {
+            min: 1,
+            allowAuto: true,
+            required: !allowIgnore,
+            alternateValues: alternateValues
```
KyleLilly, line 182:
```suggestion
            required,
            alternateValues
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -156,11 +189,27 @@ const linearIntProperties = {
...
+
+const categoricalParameterProperties = {
+    type: HyperparameterType.CATEGORICAL,
+    typeOptions: [HyperparameterType.CATEGORICAL, HyperparameterType.STATIC],
+};
+
 const continuousParameterProperties = {
     type: HyperparameterType.CONTINUOUS,
-    typeOptions: [HyperparameterType.STATIC, HyperparameterType.CONTINUOUS],
+    typeOptions: [HyperparameterType.CONTINUOUS, HyperparameterType.STATIC],
```
KyleLilly, line 209: These were previously ordered this way on purpose @dustins may remember the exact reasoning but I'm fairly certain it was related to differences between rendering the inputs on a training job form vs an hpo training job definition form. Just want to call this out so we can confirm this doesn't break anything.

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -169,284 +218,343 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
         metadata: xgboostMetadata as AlgorithmMetadata,
         defaultHyperParameters: [{
-            type: HyperparameterType.STATIC,
-            typeOptions: [HyperparameterType.STATIC],
+            ...staticParameterProperties,
             key: 'early_stopping_rounds',
             value: [],
             description: 'The model trains until the validation score stops improving. Validation error needs to decrease at least every early_stopping_rounds to continue training. SageMaker hosting uses the best model for inference.',
-            zValidator: integerIntervalValidator([0,Number.POSITIVE_INFINITY], IntervalStrategy.RIGHT_OPEN).optional(),
+            zValidator: numberValidator('learning_rate',
```
KyleLilly, line 225:
```suggestion
            zValidator: numberValidator('early_stopping_rounds',
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -169,284 +218,343 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
         }, {
-            type: HyperparameterType.INTEGER,
-            typeOptions: [HyperparameterType.INTEGER, HyperparameterType.STATIC],
+            ...integerParameterProperties,
             scalingType: 'Linear',
             key: 'num_round',
             value: [],
             description: 'The number of rounds to run the training.',
-            zValidator: integerIntervalValidator([1, Number.POSITIVE_INFINITY], IntervalStrategy.RIGHT_OPEN),
+            zValidator: numberValidator('learning_rate',
```
KyleLilly, line 241:
```suggestion
            zValidator: numberValidator('num_round',
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -169,284 +218,343 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
             description: 'Verbosity of printing messages.',
         }, {
-            type: HyperparameterType.STATIC,
-            typeOptions: [HyperparameterType.STATIC],
+            ...staticParameterProperties,
             key: 'nthread',
             value: [],
             description: 'Number of parallel threads used to run xgboost.',
-            zValidator: integerIntervalValidator([1, Number.POSITIVE_INFINITY], IntervalStrategy.RIGHT_OPEN).optional()
+            zValidator: numberValidator('learning_rate',
```
KyleLilly, line 263:
```suggestion
            zValidator: numberValidator('nthread',
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -169,284 +218,343 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
             scalingType: 'Linear',
             key: 'min_child_weight',
             value: ['1'],
             description: 'Minimum sum of instance weight (hessian) needed in a child. If the tree partition step results in a leaf node with the sum of instance weight less than min_child_weight, the building process gives up further partitioning. In linear regression models, this simply corresponds to a minimum number of instances needed in each node. The larger the algorithm, the more conservative it is.',
-            zValidator: floatIntervalValidator([0, Number.POSITIVE_INFINITY], IntervalStrategy.RIGHT_OPEN).optional()
+            zValidator: numberValidator('min_child_weight',
+                {   
+                    min: 0, 
+                    isFloat: true
+                })
```
KyleLilly, line 310: Another thing to check with the linter - you've got a combination of using a trailing comma and not using them.

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -169,284 +218,343 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
-            typeOptions: [HyperparameterType.STATIC],
+            ...staticParameterProperties,
             key: 'max_bin',
             value: ['256'],
             description: 'Maximum number of discrete bins to bucket continuous features. Used only if tree_method is set to hist.',
-            zValidator: integerIntervalValidator([1, Number.POSITIVE_INFINITY], IntervalStrategy.RIGHT_OPEN).optional()
+            zValidator: numberValidator('max_bin',
+            {   
+                min: 0, 
+            })
```
KyleLilly, line 462:
```suggestion
                {   
                    min: 0, 
                })
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -633,40 +949,41 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
             },
             {
                 key: 'min_epochs',
                 value: ['5'],
-                type: HyperparameterType.STATIC,
-                typeOptions: [HyperparameterType.STATIC],
+                ...staticParameterProperties,
                 description:
                     'The minimum number of epochs to train before early stopping logic is invoked.',
                 zValidator: positiveIntValidator('5', false),
```
KyleLilly, line 965:
```suggestion
                zValidator: positiveIntValidator('min_epochs', false),
```

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -1250,15 +1608,19 @@ export const ML_ALGORITHMS: Algorithm[] = [
                 key: 'enc0_layers',
                 value: ['auto'],
                 ...linearIntProperties,
-                zValidator: rangeValidator('enc0_layers', 1, 4, false, false, true),
+                zValidator: numberValidator('enc0_layers',
+                    {   
+                        min: 1, 
+                        max: 100,
```
KyleLilly, line 1614:
```suggestion
                        max: 4,
```
Unless this was previously incorrect.

KyleLilly (changes requested):

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -470,12 +578,213 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
+                {   
+                    min: 0, // 0 ≤ float ≤ 1
+                    max: 1,
+                    isFloat: true
+                })
+        },
+        {
+            ...categoricalParameterProperties,
+            key: 'optimizer',
+            value: ['adam','adagrad','nag','rmsprop','sgd'], // Docs say default is 'sgd', but in console all are selected
```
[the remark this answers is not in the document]
LightSeekerSC: Sounds good. I was replicating the current console which behaves in this way. We can do it the correct way according to documentation though.

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -71,83 +73,114 @@ export type Hyperparameter = {
...
-    max: number,
-    isFloat = false,
-    required = true,
-    allowAuto = false
-) => {
-    const validationMessage = `${fieldName} must be ${
-        isFloat ? 'a float' : 'an integer'
-    } value in the range [${min},${max}]${allowAuto ? ' or be \'auto\'' : ''}`;
+    // Create validation message for errors
+    const validationMessage = `${fieldName}${props.required ? ' is required and' : '' } must be ${
```
AlejandroRigau, line 142: this variable looks extremely convoluted for it to be a "1 liner". Can this be split up into multiple lines to make it more readable?

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -169,284 +218,343 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
         }, {
-            type: HyperparameterType.CONTINUOUS,
-            typeOptions: [HyperparameterType.CONTINUOUS, HyperparameterType.STATIC],
+            ...continuousParameterProperties,
             scalingType: 'Linear',
             key: 'gamma',
             value: ['0'],
             description: 'Minimum loss reduction required to make a further partition on a leaf node of the tree. The larger, the more conservative the algorithm is.',
-            zValidator: floatIntervalValidator([0,Number.POSITIVE_INFINITY], IntervalStrategy.RIGHT_OPEN).optional()
+            zValidator: numberValidator('gamma',
```
AlejandroRigau, line 285: I think all these number validators can be all in the same line. Weird that our linters did not change that. Is pre-commit installed in your dev environment? 

We would probably not want to rely on pre-commit and our github actions should be catching this.

AlejandroRigau (changes requested):
I would suggest running this through our pre-commit hooks and verify why github actions are not linting things by default.

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: Bringing this up as a team discussion. I think the best path forward is to move the new on to this validator file for re-use. Clarifying direction with the team, but yes I will remove these functions too. They shouldn't be used elsewhere.

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: This is how the console renders it and currently there is inconsistency between the type capitalizations.

Screenshot 2024-03-26 at 12 04 07 PM
Screenshot 2024-03-26 at 12 04 29 PM

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: Yeah, something may have changed? I noticed that it doesn't automatically run during the local hosting like it used to or it isn't working as expected. I will try running a build to see what the difference is.

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: No reason other than trying to stay consistent with previous. Consolidating into just alternateValues is probably preferred, so I will do that.

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: Yeah we can try. This was complex before, but I can try to make it simpler.

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: Understood. I will re-check the fields this applies to and ensure the proper usage.

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: Understood. I will check that. This is the ordering in the console which is why I switched it, but it could have adverse affects on our training rendering, so I will check.

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: I was doing it in the dev container. I am used to the `npm run start` catching these instantly which did not occur. I will try a build action.

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -470,12 +578,213 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
+        },
+        {
+            ...staticParameterProperties,
+            key: 'crop_size',
+            value: ['240'],
+            description: 'The image size for input during training. We randomly rescale the input image based on base_size, and then take a random square crop with side length equal to crop_size. The crop_size will be automatically rounded up to multiples of 8.',
+            zValidator: numberValidator('crop_size',
+                {   
+                    min: 16,
+                    includeMin: false
```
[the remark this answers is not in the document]
LightSeekerSC: Okay, I will check this. This may have been based on their documentation.

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -470,12 +578,213 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
+                    min: 1, // positive integer
+                    required: true
+                })
+        },
+        {
+            ...staticParameterProperties,
+            key: 'early_stopping_min_epochs',
+            value: ['5'],
+            description: 'The minimum number of epochs that must be run.',
+            zValidator: numberValidator('early_stopping_min_epochs') // integer
```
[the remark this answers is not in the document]
LightSeekerSC: Docs said integer, but yes, this must be >= 0. Updated

frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -470,12 +578,213 @@ export const ML_ALGORITHMS: Algorithm[] = [
...
+        },
+        {
+            ...staticParameterProperties,
+            key: 'early_stopping_patience',
+            value: ['4'],
+            description: 'The number of epochs that meet the tolerance for lower performance before the algorithm enforces an early stop.',
+            zValidator: numberValidator('early_stopping_patience') // integer
```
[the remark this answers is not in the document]
LightSeekerSC: Same issue. Docs were incorrect. Should we put in a ticket for their update?

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: Linter still doesn't seem to be working. Moving on without it for now

frontend/src/entities/jobs/algorithms.tsx
LightSeekerSC: Linter seems to not be working properly

frontend/src/entities/jobs/algorithms.tsx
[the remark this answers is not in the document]
KyleLilly: Did this turn out to be correct as written (not allowing the min)?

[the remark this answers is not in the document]
LightSeekerSC: [Documentation:
](https://docs.aws.amazon.com/sagemaker/latest/dg/segmentation-hyperparameters.html)
```
Valid values: positive integer > 16

Default value: 240
```
But yes, it does appear to be valid. Removing the min.

AlejandroRigau (approved):

KyleLilly (approved):

PR #24, opened 2024-03-28
Fix linting of frontend, address existing issues that snuck in

KyleLilly:
*Issue #, if available:*
N/A

*Description of changes:*
Adds `.eslintrc` to frontend directory so that stylistic rules are applied to frontend in addition to the CDK code. Spelling dictionary from root level `.eslintrc` file is already applied to frontend files. Confirmed this addressed the issues by running `act` locally.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,7 @@
     "parser": "@typescript-eslint/parser",
     "parserOptions": {
         "ecmaVersion": 12,
-        "sourceType": "module",
+        "sourceType": "module"
     },
     "plugins": [
         "@stylistic",
@@ -135,6 +135,7 @@
                     "debounce",
                     "debounced",
                     "decrypt",
+                    "deeplab",
                     "depthwise",
                     "deterministically",
                     "dhcp",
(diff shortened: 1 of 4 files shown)


LightSeekerSC (approved):

AlejandroRigau:
since we are adding this stuff specifically for `act`, can we also add it in our pre-commit?

Should be something like: https://github.com/pre-commit/mirrors-eslint
```
-   repo: https://github.com/pre-commit/mirrors-eslint
    rev: ''  # Use the sha / tag you want to point at
    hooks:
    -   id: eslint
```

I believe this would leverage the .eslintrc you are adding to the front end

AlejandroRigau (approved):

LightSeekerSC (approved):

KyleLilly:
> since we are adding this stuff specifically for `act`, can we also add it in our pre-commit?
> 
> Should be something like: https://github.com/pre-commit/mirrors-eslint
> 
> ```
> -   repo: https://github.com/pre-commit/mirrors-eslint
>     rev: ''  # Use the sha / tag you want to point at
>     hooks:
>     -   id: eslint
> ```
> 
> I believe this would leverage the .eslintrc you are adding to the front end

Will add in a follow-up.

PR #33, opened 2024-04-03
[AIML-ADC-5626] Autodetect batch translate fails in unsupported regions

dustins:
*Issue #, if available:* AIML-ADC-5626

*Description of changes:*

- exposed AWS_REGION from cdk `lib/constants.ts` to frontend `window.env`
- added a check to exclude 'auto' source language detection in unsupported regions

### DIFF
--- a/lib/stacks/api/restApi.ts
+++ b/lib/stacks/api/restApi.ts
@@ -35,25 +35,9 @@ import { LogGroup } from 'aws-cdk-lib/aws-logs';
 import { Bucket } from 'aws-cdk-lib/aws-s3';
 import { BucketDeployment, Source } from 'aws-cdk-lib/aws-s3-deployment';
 import { StringParameter } from 'aws-cdk-lib/aws-ssm';
-import {
-    ADDITIONAL_LAMBDA_ENVIRONMENT_VARS,
-    APPLICATION_NAME,
-    BUCKET_DEPLOYMENT_ROLE_ARN,
-    COMMON_LAYER_ARN_PARAM,
-    ENABLE_ACCESS_LOGGING,
-    ENABLE_GROUNDTRUTH,
-    IDP_ENDPOINT_SSM_PARAM,
-    INTERNAL_OIDC_URL,
-    LAMBDA_ARCHITECTURE,
-    LAMBDA_RUNTIME,
-    MANAGE_IAM_ROLES,
-    OIDC_CLIENT_NAME,
-    OIDC_REDIRECT_URI,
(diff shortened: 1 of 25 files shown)


lib/stacks/api/restApi.ts
```diff
@@ -305,7 +306,8 @@ export class RestApiStack extends Stack {
             ENABLE_GROUNDTRUTH: ENABLE_GROUNDTRUTH,
             SYSTEM_BANNER: props.systemBannerConfiguration,
             APPLICATION_NAME: APPLICATION_NAME,
-            DATASET_BUCKET: props.dataBucketName
+            DATASET_BUCKET: props.dataBucketName,
+            AWS_REGION,
```
AlejandroRigau, line 310: should this be

```suggestion
AWS_REGION: AWS_REGION,
```

this?

KyleLilly: Not totally necessary in this case. If the property name and the variable you're assigning are the same you can omit the explicit assignment. Admittedly this isn't totally consistent across the codebase.

KyleLilly (dismissed):

lib/stacks/api/restApi.ts
dustins: We aren't consistent about this throughout the project and I tend to use it when I can but I'm going to accept this suggestion to keep consistency in this file.

AlejandroRigau (dismissed):

frontend/src/entities/batch-translate/create/batch-translate-create.tsx
```diff
@@ -76,6 +77,13 @@ export function BatchTranslateCreate () {
     scrollToPageHeader();
     DocTitle('Create translation job');
 
+    // the dependencies needed for language detection aren't available in us-iso-east-1
+    const supportsAutoLanguageDetection = window.env.AWS_REGION !== 'us-iso-east-1';
```
douglas1850, line 81: If Translate is ever built into us-isob-east-1 or any other region where auto isn't supported it would be nice if we made this more extensible. 

Maybe we have a list defined here or in a constants/util class which defines all regions auto isn't supported in, and then we check if the user's region is in the list of unsupported regions:

```
const supportsAutoLanguageDetection = !autoSupportedRegions.includes(window.env.AWS_REGION);
```

LightSeekerSC: I agree that this should be a list of unavailable regions. I think the inclusive list would be much larger

PR #38, opened 2024-04-04
[AIML-ADC-7397] add ability for custom branding on docs

AlejandroRigau:
*Issue #, if available:* AIML-ADC-7397

*Description of changes:*
- Add ability for custom branding on docs. This is done by passing a param through out the .md files which loads the application name from the constants.ts
- Changed a few titles to better fit the new use of `{{ $params.APPLICATION_NAME }}`. This particular interpolation cannot be used in titles. 

Screenshot 2024-04-04 at 6 31 30 PM


LightSeekerSC (approved):
Assuming this runs during build-time, looks good to me. Was this verified with a deployment? Approving assuming this was tested.

AlejandroRigau:
I tested this by doing two main things:
- `npm run docs:dev` which run dev server and I can see the changes there. I went through all pages looking for things that made sense to replace. The only page where I did not change this was in the branding.md page. 
- `npm run docs:build` and `npm run docs:preview` also lets me simulate the production deployment which it looked great and the same as docs:dev
- I had to modify some of the headers since it seems that headers dont like the {{ variable }} syntax and it would affect how the search would work

KyleLilly (approved):

PR #41, opened 2024-04-08
Cypress Integration Tests

LightSeekerSC:
*Issue #, if available:*

*Description of changes:*
- Adds historic integration tests
- Adds integration test compatibility with Cognito and the new landing page
- Updates integration tests for new UI / functionality changes


cypress/cypress/screenshots/batch-translate.spec.cy.ts/Batch Translate Tests -- Create Batch Translate Job -- before all hook (failed).png
KyleLilly, line 1: Did you mean to include screenshots of failed tests? The parent directory should be in `.gitignore`.

KyleLilly (changes requested):

cypress/cypress/screenshots/batch-translate.spec.cy.ts/Batch Translate Tests -- Create Batch Translate Job -- before all hook (failed).png
LightSeekerSC: I did not mean to, no

cypress/cypress/fixtures/example.json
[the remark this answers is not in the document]
LightSeekerSC: I don't know, but we can get rid of it

cypress/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "husky": "^8.0.0",
+        "lint-staged": "^13.0.3",
+        "typescript": "^4.8.2"
+    },
+    "scripts": {
+        "cypress:open": "cypress open",
+        "cypress:run": "cypress run",
+        "clean": "rm -rf build/ node_modules/",
+        "lint:fix": "eslint --fix cypress/",
+        "release": "cp -R node_modules build && cp -R cypress build && cp cypress.config.ts build && cp tsconfig.json build && cp cypress.d.ts build && cp package.json build && cp package-lock.json build"
```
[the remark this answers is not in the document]
LightSeekerSC: Definitely just a hold-over. Probably not needed

cypress/package.json
```diff
@@ -0,0 +1,43 @@
+{
```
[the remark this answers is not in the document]
LightSeekerSC: Not sure if we need a separate package.json. Are you implying we should add commands to the main package.json that allow running tests from the root directory?

cypress/tsconfig.json
```diff
@@ -0,0 +1,34 @@
+{
```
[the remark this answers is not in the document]
LightSeekerSC: Not sure. I can try removing.

cypress/package.json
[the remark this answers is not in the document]
LightSeekerSC: I think at this time I am going to lean towards keeping them separate. While we haven't decided if cypress is our permanent path forward, I would rather not consolidate as it will be harder to identify all the components to remove if we choose a framework other than cypress for the long-term.

cypress/tsconfig.json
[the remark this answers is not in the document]
LightSeekerSC: Same answer as for package.json. Until we decide on a preferred testing framework, I think it is better to keep these separate.

cypress/cypress/e2e/batch-translate.spec.cy.ts
```diff
@@ -37,14 +37,12 @@ describe('Batch Translate Tests', () => {
         ]
     };
 
-    /* eslint-disable spellcheck/spell-checker */
     const knownAccessibilityDefects = [
         'The aria-posinset attribute is not allowed on this element because it does not have the correct role attribute',
         'The aria-posinset attribute is not allowed on this LI because it does not have the correct role attribute',
         'The aria-posinset and aria-setsize attribute pair is not allowed on this LI because it does not have the correct role attribute',
         'The aria-setsize attribute is not allowed on this LI because it does not have the correct role attribute',
     ];
```
KyleLilly, line 45: Let's remove these all together - they're no longer used.

cypress/cypress/support/auth-provider-commands/keycloak.ts
```diff
@@ -40,7 +39,7 @@ const loginToKeycloak = (baseUrl: string, username: string, password: string) =>
         {
             validate: () => {
                 cy.wrap(sessionStorage)
-                    .invoke('getItem', 'oidc.user:https://ec2-52-55-68-28.compute-1.amazonaws.com/realms/mlspace:web-client')
+                    .invoke('getItem', `oidc.user:${keyCloakUrl}realms/mlspace:web-client`)
```
KyleLilly, line 42: Realm and client should probably be configurable similar to URL. Doesn't need to change at the moment though as we're not currently including keycloak with deployments.

cypress/cypress/support/test-initializer/types.ts
```diff
@@ -0,0 +1,27 @@
+enum AuthType {
+    Cognito = 'COGNITO',
+    Idp = 'IDP',
```
KyleLilly, line 3: This is a little weird because `IdP` stands for Identity Provider of which Cognito would be one option. Assuming `Idp` in this case is the mock Idp then we should probably just remove it - the mock IdP isn't something that will be used moving forward and isn't something that is of use external to our team. The same comment aplies to the `mockIdp.ts` provider and any other references in this PR.

KyleLilly:
The Apache license header needs to be added to all these files.

.devcontainer/post_create_command.sh
```diff
@@ -14,4 +14,8 @@ pip3 install -r test/requirements.txt
 
 cd ..
 
+# Setup cypress
+sudo apt-get install -y libgtk2.0-0 libgtk-3-0 libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb libgbm-dev
```
AlejandroRigau, line 18: we should not add cypress this way. We should add it in the features section above. An easy way to do this is to go to vscode and do CMD + SHFT + P and type `Configure Container Features` where you can search for Cypress.

also, if the package.json already includes cypress, we dont need to `npm install cypress` since the other npm installs would install it to node_modules.

cypress/package.json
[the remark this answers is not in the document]
AlejandroRigau: I think we should be adding cypress to the frontend package.json as a devDependency. That way we can make sure that our testing framework is aligned and using the same versions as our frontend for things like cloudscape, eslint, or even typescript.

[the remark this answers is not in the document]
AlejandroRigau: if we are holding off until we can determine if we want to continue using cypress at all, then maybe we should decide this before pushing these changes to main.

AlejandroRigau (changes requested):

.devcontainer/post_create_command.sh
AlejandroRigau: > We should add it in the features section above.
I meant to say that we should add it in the `devcontainer.json`. Thats where we control what things are installed in the dev container. Post_create_command is just something we run after the container has been built but we can set it up so that the container already has cypress installed rather than installing it ourselves

LightSeekerSC:
Don't review yet. Verifying dev container fix

LightSeekerSC:
Dev container has been reviewed and this should be ready for review again

.devcontainer/post_create_command.sh
LightSeekerSC: I wasn't able to use the features. The only method that worked was post_create_commands with some special tweaks

AlejandroRigau (approved):

cypress/cypress/support/test-initializer/types.ts
KyleLilly: I don't really think it makes sense as that's not something that will be maintained or available externally.

KyleLilly (approved):
Generally I would still prefer we get rid of the extra package-lock and package json now that we've got a mono repo and I also don't think there is any value keeping the mock IdP stuff at this point but I don't want to hold this up any longer.

PR #42, opened 2024-04-09
Fixes HPO Training Job Definition Errors

LightSeekerSC:
*Issue #, if available:* https://issues.amazon.com/issues/AIML-ADC-7542

*Description of changes:*
- Creates a notification in the case that there are form errors when submitting an HPO training definition (previously would just not submit, but issue was not visible to user)
- Fixes issue where removing a metric definition would cause form errors

Error Notification:
Recording 2024-04-09 at 18 56 49

Fixed Metric Definition Form Error:
Recording 2024-04-09 at 18 58 05


KyleLilly:
Is there anything preventing us from catching the issue on `next` click as opposed to waiting until they are on step 3 of the wizard and actually submitting the job?

LightSeekerSC:
> Is there anything preventing us from catching the issue on `next` click as opposed to waiting until they are on step 3 of the wizard and actually submitting the job?

In theory no. The fact that we have separate form validation at the 3rd stage that is evaluating the entirety of the form is part of the problem. In theory, I would think that each section of the wizard should be validating itself. That being said, the "form error" in this case was more of a bug overall. This is an issue that training jobs also face and the error isn't surfaced there either as the field for which the error is occurring doesn't exist and therefore isn't displayed.

In theory you are correct that yes, form errors should be caught on the relevant page.

KyleLilly (approved):

AlejandroRigau (approved):

PR #44, opened 2024-04-11
[AIML-ADC-8090] Fix issue with fetching specific projects during report generation

KyleLilly:
*Issue #, if available:*
AIML-ADC-8090

*Description of changes:*
Updated `get_all` method in project dao to not short circuit if `include_suspended` was set to `True`. While generating reports we will always include suspended projects but the existing logic prevented us from only including specific projects in the report as the ddb scan was being executed without a filter. The updated body will apply a filter for `suspended = false` if `include_suspended` is `False` (default), it will then add an additional project filter if present, and will then execute the scan. The existing unit test was incorrectly asserting that `get_all` with no arguments returned 2 entries when the seeded projects include 1 suspended and 1 active project. Since `include_suspended` defaults to `False` that test should only expect a single matching project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # MLSpace
 
+[![Full Documentation](https://img.shields.io/badge/Full%20Documentation-blue?style=for-the-badge&logo=Vite&logoColor=white)](https://awslabs.github.io/mlspace/)
+
 ## Deployment Prerequisites
 
 ### Software
--- a/backend/src/ml_space_lambda/data_access_objects/project.py
+++ b/backend/src/ml_space_lambda/data_access_objects/project.py
@@ -139,9 +139,14 @@ class ProjectDAO(DynamoDBObjectStore):
     def get_all(
         self, include_suspended: Optional[bool] = False, project_names: Optional[List[str]] = None
     ) -> List[ProjectModel]:
-        if include_suspended:
-            json_response = self._scan().records
-        elif project_names is not None:
+        expression_attribute_values: Dict[str, Any] = {}
+        filter_expressions = []
+
+        if not include_suspended:
+            expression_attribute_values[":suspended"] = include_suspended
+            filter_expressions.append("suspended = :suspended")
+
+        if project_names is not None:
             # If we specified a filter but it was empty then just return an empty list
             if not project_names:
                 return []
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,5 +1,7 @@
 # MLSpace
 
+[![Full Documentation](https://img.shields.io/badge/Full%20Documentation-blue?style=for-the-badge&logo=Vite&logoColor=white)](https://awslabs.github.io/mlspace/)
```
AlejandroRigau, line 3: Where does this documentation get built from?

backend/src/ml_space_lambda/data_access_objects/project.py
```diff
@@ -154,15 +159,16 @@ def get_all(
...
 
-            json_response = self._scan(
-                filter_expression=f"suspended = :suspended AND #name IN ({', '.join(expressions)})",
-                expression_names={"#name": "name"},
-                expression_values=json.loads(dynamodb_json.dumps(expression_attribute_values)),
-            ).records
-        else:
-            json_response = self._scan().records
+        json_response = self._scan(
+            filter_expression=" AND ".join(filter_expressions) if filter_expressions else None,
```
AlejandroRigau, line 168: Do we still need this AND here?

README.md
KyleLilly: This [github action on merge](https://github.com/awslabs/mlspace/blob/main/.github/workflows/deploy.yml). Previous runs [here](https://github.com/awslabs/mlspace/actions/workflows/deploy.yml).

backend/src/ml_space_lambda/data_access_objects/project.py
KyleLilly: Yes - if there are multiple filter expressions they need to be joined with `AND` in this particular case. Options are:

- No filter expression in which case this is `None`
- Suspension only filter expression `suspended = :suspended`
- Specific project filter expression `#name IN (<list of names>)`
- Suspension and specific project filter expression `suspended = :suspended AND #name IN (<list of names>)`

Using `" AND ".join(filter_expressions)` lets us handle all of those cases.

AlejandroRigau (approved):

LightSeekerSC (approved):

PR #49, opened 2024-04-15
Addresses Cypress Resizer Errors

LightSeekerSC:
*Issue #, if available:* https://issues.amazon.com/issues/AIML-ADC-5792

*Description of changes:* 
- Prevents resizer errors from failing tests
- Enables Cypress to run on a localhost execution of Cypress

Recording 2024-04-12 at 17 00 32


douglas1850 (approved):

cypress/cypress.config.ts
```diff
@@ -29,6 +29,7 @@ module.exports = defineConfig({
   env: {
     auth_type: AuthType.Cognito,
     base_url: "",
+    lambda_endpoint: "",
```
KyleLilly, line 32: Can you add some documentation on this parameter and when you'd use it (running UI locally against deployed API GW for lambdas)?

cypress/cypress/support/project-commands/project.ts
```diff
@@ -68,7 +66,7 @@ const deleteProjectsWithPrefix = (projectNamePrefix: string, maxNumberOfDeletes
         if (sessionStorageKey.startsWith('oidc.user')) {
             // Use OIDC header to connect to the project API
             cy.request({
-                url: `${BASE_URL}/project/`,
+                url: LAMBDA_ENDPOINT ? `${LAMBDA_ENDPOINT}/project/` : `${BASE_URL}/project/`,
```
KyleLilly, line 69: Is this the only place we need this change? We probably need a comment on that constant explaining when you should use `LAMBDA_ENDPOINT` vs `BASE_URL`

AlejandroRigau:
LGTM after addressing Kyle's comments.

cypress/cypress/support/auth-provider-commands/cognito.ts
```diff
@@ -35,8 +35,8 @@ const loginToCognito = (baseUrl: string, username: string, password: string) =>
         () => {
             // Handle cognito portal information
             cy.request(baseUrl + '/env.js').then((resp) => {
-                const OIDC_URL_REGEX = /["']OIDC_URL['"]:\s*['"]([A-Za-z:\-._/0-9]+)['"]/;
-                const OIDC_APP_NAME_REGEX = /["']OIDC_CLIENT_NAME['"]:\s*['"]([A-Za-z:\-._/0-9]+)['"]/;
+                const OIDC_URL_REGEX = /["']*OIDC_URL['"]*:\s*['"]*([A-Za-z:\-._/0-9]+)['"]*/;
+                const OIDC_APP_NAME_REGEX = /["']*OIDC_CLIENT_NAME['"]*:\s*['"]*([A-Za-z:\-._/0-9]+)['"]*/;
```
dustins, line 39: FWIW I think this should probably be `?` instead of `*` right? We wouldn't want to match (and it wouldn't be valid) to have >1 quote. But this works so I'd say not necessary to change it unless you're already making other changes.

dustins (approved):

KyleLilly (approved):

douglas1850 (approved):

PR #50, opened 2024-04-15
[AnIMaL-25962] [MLSpace] Implement polling to update resource status

AlejandroRigau:
*Issue #, if available:* AnIMaL-25962

*Description of changes:*
- Resource statuses in tables refresh every 60 seconds to reflect current status
- When creating or stopping a resource from the list view the status will be eventually reflected in the table without user intervention
- When manipulating a resource from the details view the status will be accurately reflected in the details page as it changes


AlejandroRigau:
I was able to test it with most of the details pages and tables. I will continue doing some testing while this gets reviewed to ensure stability.

frontend/src/shared/util/hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+    const [isBackgroundRefreshing, setIsBackgroundRefreshing] = useState(false);
+    const callbackAction = useCallback(action, [action, ...deps]);
+    
+    useEffect(() => {
+        const timerId = setInterval(() => {
+            if (condition) {
+                setIsBackgroundRefreshing(true);
+                callbackAction();
+            }
+        }, 60000);
```
dustins, line 29: Any possible reason we might want this configurable? I could see a fourth parameter that could control the interval duration.

frontend/src/shared/util/hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import { useCallback, useEffect, useState } from 'react';
+
+export function useBackgroundRefresh (action: () => void, deps: readonly unknown[] = [], condition = true): boolean {
```
dustins, line 19: Some nit picks. I'd love to see a docblock comment explaining this function, giving an example. The documentation I'm sure would explain it but because condition is such a generic name I am not really sure what I should be using that for. I'd maybe change it to `shouldRefresh` or `needsRefresh` maybe?

frontend/src/shared/util/hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+    const [isBackgroundRefreshing, setIsBackgroundRefreshing] = useState(false);
+    const callbackAction = useCallback(action, [action, ...deps]);
+    
+    useEffect(() => {
+        const timerId = setInterval(() => {
+            if (condition) {
```
dustins, line 25: Should we put the condition around everything inside useEffect? That way we could skip running the interval repeatedly when `condition` is false and doing nothing right?

AlejandroRigau: Good point. Ill change this.

frontend/src/shared/util/hooks.ts
AlejandroRigau: I personally don't see the need to make it configurable but maybe @KyleLilly has more context on what would be more beneficial for the customer. We could potentially make this configurable in constants.ts or similar config file.

dustins: It was just an idea because this is a generic hook I could  see it being used other places that might want something different. But if you don't feel there is a need we can let it stand as-is. It would be super easy in the future to change it if is needed.

KyleLilly: Yeah - this should really be configurable. Something that should probably be tied in with the in-app configuration work being done @douglas1850 .

KyleLilly (dismissed):

frontend/src/shared/util/hooks.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function useBackgroundRefresh (action: () => void, deps: readonly unknown[] = [], condition = true): boolean {
+    /**
+     * A custom hook that runs an action periodically in the background to refresh data.
+     * 
+     * @param {Function} action - The function to run periodically to refresh data 
+     * @param {Array} deps - The dependencies of the action function
+     * @param {boolean} condition - A condition that must be true for the action to run
+     * @returns {boolean} isBackgroundRefreshing - Whether the background refresh is currently running
+    */
```
dustins, line 27: Excellent comment but the docblock should be above the function. I think Python is the only weird language that I know of that comments inside the function.

AlejandroRigau: great call-out. Ill fix this.

dustins (approved):

frontend/src/entities/emr/detail/emr-clusters-detail.tsx
```diff
@@ -63,6 +64,11 @@ function EMRDetail () {
         }
     }, [dispatch, basePath, projectName, clusterId, clusterName]);
 
+    // Refresh data in the background to keep state fresh
+    const isBackgroundRefreshing = useBackgroundRefresh(() => {
```
douglas1850, line 68: Just my own preference, but I don't think we should be doing background refreshes on entities which are not prone to frequent updates. 

I think auto-refreshing on jobs (training, translate, transform) is great and that you've implemented this in a very nice way, but resources like EMR and Notebooks change state very infrequently and my concern would be that if someone left the EMR/Notebook detail view open and then left their computer, they'd run up a bill from these repeated describe calls. 

Out of scope for this PR/task would be adding a little refresh button to the Notebook and EMR pages, whereas for jobs it's nice that we check for their status on an interval and then have a definite stop condition when the job has reached a terminal state.

KyleLilly: This is similar to the AWS console. While technically in the console you aren't charged the cost in MLSpace would be negligible - $0.2583 per million background refreshes based on ddb read and lambda invocations (prices based on us-east-2). If customers are concerned they can increase the frequency of refreshes. This feature is in direct response to a feature request from multiple customers.

KyleLilly (approved):

PR #52, opened 2024-04-16
Metric Definition Duplicates

LightSeekerSC:
*Issue #, if available:* https://issues.amazon.com/issues/AIML-ADC-6788

*Description of changes:*
- Adds a check to ensure that when built-in containers are used for BYOM, duplicate metrics aren't created
- Updated unit tests to work with new testing

ORIGINAL HPO JOB:
Screenshot 2024-04-16 at 5 16 02 PM

COPIED HPO JOB:
Screenshot 2024-04-16 at 5 16 18 PM


backend/src/ml_space_lambda/utils/image_uris.py
```diff
@@ -31,6 +31,46 @@
 
 logger = logging.getLogger(__name__)
 
+### START CUSTOM MLSPACE ADDITIONS FOR DETECTING BUILT-IN ALGORITHMS
+# Most images are named the same as their frameworks except for the below exceptions
+IMAGE_TO_FRAMEWORK_NAME_CONVERSIONS = {"sagemaker-xgboost": "xgboost"}
```
KyleLilly, line 36: It doesn't look like any part of this custom addition actually relies on internal methods in this file and only leverages `retrieve` which we could call externally. Given that this file is basically a copy from the SageMaker SDK I think it'd be better to keep this in it's own utility file (or just outside of this file). When OCI container support is available in region we'll be able to bundle the SageMaker SDK directly.

backend/test/conftest.py
```diff
@@ -23,6 +24,10 @@
 from ml_space_lambda.enums import DatasetType
 
 
+def pytest_generate_tests(metafunc):
+    os.environ["LAMBDA_TASK_ROOT"] = "./src/"
```
KyleLilly, line 28: Can you add a comment about why this is needed?

backend/src/ml_space_lambda/training_job/lambda_functions.py
```diff
@@ -36,6 +42,7 @@
 
 @api_wrapper
 def create(event, context):
+    deleted_metric_definitions = False
```
douglas1850, line 45: Do we need this when we assign it a value on line 107?

LightSeekerSC: This is being set to true in two different places and then assigns the value once. We could just remove this and then assign the value twice.

backend/src/ml_space_lambda/training_job/lambda_functions.py
```diff
@@ -96,16 +103,23 @@ def create(event, context):
...
         if "You can't override the metric definitions for Amazon SageMaker algorithms" in error.response["Error"]["Message"]:
             del training_job_definition["AlgorithmSpecification"]["MetricDefinitions"]
             response = sagemaker.create_training_job(**training_job_definition)
-            response["DeletedMetricsDefinitions"] = True
+            deleted_metric_definitions = True
         else:
             raise error
 
+    # If the metric definitions were deleted due to the job being submited as a BYOM, but the algorithm was a built-in, add a response showing the metrics definitions were removed
+    if deleted_metric_definitions:
```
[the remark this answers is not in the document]
LightSeekerSC: In this case, the thought was that we probably don't want to assign this value if it is false as it is being sent in a network response. I think the preferred values here are `True` or undefined

AlejandroRigau:
LGTM after all other conversations are resolved

backend/test/conftest.py
KyleLilly: That's fine for and probably makes sense for things like the region etc. I was more interested in why we had to set the `LAMBDA_TASK_ROUTE` as that's new with this PR.

LightSeekerSC: It was needed because the Sagemaker SDK check invoked with `retrieve` checks file names and it was unable to get the env needed for locating the file names

KyleLilly: Can you add the comment in code so if someone sees it they'll know why it's there. Unlike some other env variables it's not totally clear why you'd need this.

KyleLilly (approved):

dustins (approved):

PR #54, opened 2024-04-17
Provides Greater Visibility For Additional READMEs

LightSeekerSC:
*Description of changes:* Provides greater visibility for additional READMEs within the project. Not all of our separate components have additional READMEs, so this helps clarify for users and devs which components have separate additional documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 [![Full Documentation](https://img.shields.io/badge/Full%20Documentation-blue?style=for-the-badge&logo=Vite&logoColor=white)](https://awslabs.github.io/mlspace/)
 
+## Additional READMEs
+- [Backend](backend/README.md)
+- [Cypress Integration Tests](cypress/README.md)
+
 ## Deployment Prerequisites
 
 ### Software


README.md
```diff
@@ -2,6 +2,10 @@
 
 [![Full Documentation](https://img.shields.io/badge/Full%20Documentation-blue?style=for-the-badge&logo=Vite&logoColor=white)](https://awslabs.github.io/mlspace/)
 
+## Additional READMEs
```
AlejandroRigau, line 5: I think this primary README is intended for customers rather than devs. I think these two README are more geared towards devs. I would either put this at the bottom or just add it to our internal dev wiki

KyleLilly: I don't feel super strongly about including these in general. I took a quick look at a few popular open source projects that have nested READMEs and none of them have this but if we think there's an issue with discovery I'm fine with it. Related to the comment on outside contributors that usually falls into a `CONTRIBUTING.md` or `DEVELOPMENT.md` or similar. Github does a pretty good job of surfacing the README's as you navigate the project.

LightSeekerSC: If that is standard open-source practice then I am fine with closing this as it is probably just a reflection on my experience. I will go ahead and close this for now

PR #56, opened 2024-04-18
Fixed error when no query string parameters are sent when listing Dataset files

dustins:
*Issue #, if available:*

*Description of changes:*

An error would occur when fetching the files for a Dataset with no query string parameters present. This happened because APIGW sets the `queryStringParameters` property on the `event` parameter to `None` instead of just not setting it. This causes `event.get("queryStringParameters", {})` to return `None` instead of `{}`.

### DIFF
--- a/backend/src/ml_space_lambda/dataset/lambda_functions.py
+++ b/backend/src/ml_space_lambda/dataset/lambda_functions.py
@@ -232,7 +232,7 @@ def list_resources(event, context):
 @api_wrapper
 def list_files(event, context):
     env_variables = get_environment_variables()
-    query_string_parameters = event.get("queryStringParameters", {})
+    query_string_parameters = event.get("queryStringParameters") or {}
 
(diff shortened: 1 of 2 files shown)


backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -232,7 +232,7 @@ def list_resources(event, context):
 @api_wrapper
 def list_files(event, context):
     env_variables = get_environment_variables()
-    query_string_parameters = event.get("queryStringParameters", {})
+    query_string_parameters = event.get("queryStringParameters") or {}
```
KyleLilly, line 235: Can you add a test specifically for this case?

AlejandroRigau (dismissed):

KyleLilly (approved):

AlejandroRigau (approved):

PR #57, opened 2024-04-18
Dataset Fixes/Enhancements

dustins:
*Issue #, if available:*

*Description of changes:*

Fixes for the new DatasetBrowser

- (frontend) selected items weren't being cleared when changing view contexts
- (frontend) URI for querying Dataset files was built using `window.env.LAMBDA_ENDPOINT` instead of relying on Axios to provide the base URL


frontend/src/modules/dataset/dataset-browser.reducer.tsx
```diff
@@ -173,25 +173,25 @@ export const getDatasetContents = createAsyncThunk(
...
+            searchParams.set('projectName', props.projectName);
         }
 
         if (datasetContext?.location) {
-            requestUrl.searchParams.set('prefix', datasetContext.location);
+            searchParams.set('prefix', datasetContext.location);
         }
 
-        return axios.get<ListFilesResponse>(requestUrl.toString()).then((response) => {
+        return axios.get<ListFilesResponse>(`/dataset/${scope}/${datasetContext.name}/files?${searchParams.toString()}`).then((response) => {
```
KyleLilly, line 194: Is this going to be a problem when running the UI locally?

dustins: I don't think so. I thought it was which is why I originally was making the URL using the `window.env.LAMBDA_ENDPOINT` variable but looking at the rest of our project [it](https://github.com/awslabs/mlspace/blob/main/frontend/src/entities/emr/emr.reducer.ts#L54) [really](https://github.com/awslabs/mlspace/blob/main/frontend/src/entities/notebook/notebook.reducer.ts#L55) [varies](https://github.com/awslabs/mlspace/blob/main/frontend/src/entities/user/user.reducer.ts#L52) [what](https://github.com/awslabs/mlspace/blob/main/frontend/src/entities/jobs/training/training-job.reducer.tsx#L65) we use. 

I think Axios must be handling the discrepancy for us when you give it paths which is why all the other examples still work but it failed when I provided a full URL.

Even in the right next to each other in the [same file](https://github.com/awslabs/mlspace/blob/main/frontend/src/entities/user/user.reducer.ts#L57-L71) we do it different ways.

```typescript
export const addUsersToProject = createAsyncThunk(
    'user/add_users_to_project',
    async (data: UserProjectData) => {
        const requestUrl = `project/${data.projectName}/users`;
        return axios.post<IUser>(requestUrl, { usernames: data.usernames });
    }
);

export const removeUserFromProject = createAsyncThunk(
    'user/remove_user_from_project',
    async (data: IProjectUser) => {
        const requestUrl = `/project/${data.project}/users/${data.user}`;
        return axios.delete(requestUrl);
    }
);
```

AlejandroRigau (approved):

frontend/src/modules/dataset/dataset-browser.reducer.tsx
KyleLilly: Ah yeah we've got it in the [base axios config](https://github.com/awslabs/mlspace/blob/main/frontend/src/shared/util/axios-utils.ts#L52). And then axios is handling the `/` although it would be nice if we were consistent based on the examples you linked. Either way unrelated to this so nothing to change.

KyleLilly (approved):

PR #59, opened 2024-04-19
CDK and Backend groundwork

douglas1850:
*Description of changes:* This PR lays the groundwork for the upcoming in-app configuration feature.

**CDK Changes**
* Provision a new table for storing app configurations. Version IDs are the sort key so we can easily retrieve the latest one (more on that below with ScanForwardIndex)
* Leverage `AwsCustomResource` to automatically populate the app config table with a default configuration after initial deployment
* Created 2 new APIs for the app-config path; `update_configuration` and `get_configuraion`
* Updated API Gateway to have `MethodOptions` for the app-config path so that we cache the results from `get_configuration`. NOTE: when I do the frontend changes next, I plan to make it so that when an admin submits a request to update the config, they invalidate the cache so there's no risk of users receiving an outdated config

**Backend Changes**
* Created a new lambda function for `update_configuration` and `get_configuraion`
* Created DAO for parsing the requested configuration into a configuration object
* When making the `put_item` call to DDB, use ConditionalKeyExpression to ensure that this new versionId doesn't exist for this configScope yet. If it does, then this request was modifying an outdated config and we should reject it. If it doesn't exist then this is an update on the latest config, so complete the write request
* added the `ScanForwardIndex` property to our `query` calls to DDB; this value is `True` by default with the DDB query API, so in our lambda code I've defaulted it to True as well. For the config calls I'm setting it to false so that configs are returned in descending order, with the latest versionId coming first. This way when we query for just one config, we'll always get the latest one for the given configScope
* the authorizer has been updated such that we'll allow requests from admins for updating the app-wide config (and denying everyone else) and all users will be able to GET the app-wide config. Then for projects only the admins and project owners will be able to update the project-wide config, with only project-users being able to GET the config for that project


backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -0,0 +1,62 @@
...
+
+log = logging.getLogger(__name__)
+app_configuration_dao = AppConfigurationDAO()
+
+
+@api_wrapper
+def update_configuration(event, context):
+    request = json.loads(event["body"])
+    configScope = request["configScope"]
+    version_id = request["versionId"] + 1  # increment so this will be the latest version
```
AlejandroRigau, line 31: why does it plus 1 to the requested version? If I ask for version 1, will I get version 2?

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -0,0 +1,382 @@
...
+from venv import logger
+
+from dynamodb_json import json_util as dynamodb_json
+
+from ml_space_lambda.data_access_objects.dynamo_data_store import DynamoDBObjectStore
+from ml_space_lambda.enums import ResourceType
+from ml_space_lambda.utils.mlspace_config import get_environment_variables
+
+
+class AppConfigurationModel:
```
AlejandroRigau, line 30: Would using a Type be more appropriate than using a Class? I would think that if using a Class we would have specific methods for this class but it seems we just store values here. It would also get rid of those to_dict methods I think. Same thing applies to all the other Classes in this file

I know you are probably modeling this based on the other DAO so I guess its open to discussion

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -0,0 +1,382 @@
...
+    def from_dict(dict_object: dict) -> ServiceInstanceTypes:
+        return ServiceInstanceTypes(
+            dict_object[ResourceType.NOTEBOOK.value],
+            dict_object[ResourceType.ENDPOINT.value],
+            dict_object[ResourceType.TRAINING_JOB.value],
+            dict_object[ResourceType.TRANSFORM_JOB.value],
+        )
+
+
+class EnabledServices:
```
AlejandroRigau, line 138: Same as above. I think a Type would just be cleaner but I am ok being wrong

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -0,0 +1,382 @@
...
+
+    @staticmethod
+    def from_dict(dict_object: dict) -> ProjectCreation:
+        return ProjectCreation(
+            dict_object["AdminOnly"],
+            dict_object["AllowedGroups"],
+        )
+
+
+class SystemBanner:
```
AlejandroRigau, line 337: I feel like some of these can be Types that are shared with the frontend. Benefits of a monorepo

lib/stacks/api/appConfiguration.ts
```diff
@@ -0,0 +1,77 @@
...
+        const restApi = RestApi.fromRestApiAttributes(this, 'RestApi', {
+            restApiId: props.restApiId,
+            rootResourceId: props.rootResourceId,
+        });
+
+        const apis: MLSpacePythonLambdaFunction[] = [
+            {
+                name: 'get_configuration',
+                resource: 'app_configuration',
+                description: 'Get the requested number of MLSpace application configurations, starting from the most recent',
```
AlejandroRigau, line 47: is there a reason we would request 2 app configurations at the same time? or is this description inaccurate? I figured we would always be requesting the latest one

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -0,0 +1,382 @@
...
+            "changedBy": self.changed_by,
+            "changeReason": self.change_reason,
+        }
+
+    @staticmethod
+    def from_dict(dict_object: dict) -> AppConfigurationModel:
+        return AppConfigurationModel(
+            dict_object["configScope"],
+            dict_object["versionId"],
+            dict_object["configuration"],
```
douglas1850, line 59: This should be `SettingsModel.from_dict(dict_object["configuration"]),`, changing in next commit

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -0,0 +1,62 @@
+#
```
AlejandroRigau, line 1: unrelated to this file: will this mean that users will be retrieveing the app configuration every time the page loads? Does this add significant initial load time? Seems unnecessary 

I am mostly wondering why is this not just a config file that gets dropped in the static site bucket every time we have an update in configuration? why the need for an API?

lib/stacks/infra/core.ts
```diff
@@ -410,6 +412,29 @@ export class CoreStack extends Stack {
             projectionType: ProjectionType.ALL,
         });
 
+        // App Configuration Table
+        const appConfigTable = new Table(this, 'mlspace-ddb-app-configuration', {
```
AlejandroRigau, line 416: why are we using a DDB table rather than using something like SSM Parameters? Why do we need config versioning at all?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -0,0 +1,62 @@
...
+
+@api_wrapper
+def update_configuration(event, context):
+    request = json.loads(event["body"])
+    configScope = request["configScope"]
+    version_id = request["versionId"] + 1  # increment so this will be the latest version
+
+    new_configuration = SettingsModel.from_dict(request["configuration"])
+
+    app_configuration = AppConfigurationModel(
```
AlejandroRigau, line 35: Should there also be a "Date_Changed"? That could be useful for users/Admins.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -0,0 +1,62 @@
...
+
+from ml_space_lambda.data_access_objects.app_configuration import AppConfigurationDAO, AppConfigurationModel, SettingsModel
+from ml_space_lambda.utils.common_functions import api_wrapper
+
+log = logging.getLogger(__name__)
+app_configuration_dao = AppConfigurationDAO()
+
+
+@api_wrapper
+def update_configuration(event, context):
```
AlejandroRigau, line 28: Where do users call this api from? Are we going to have some kind of admin page in MLSpace where this can be edited?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
douglas1850: This is updating the configuration. On the frontend they'll be modifying the current latest configuration (versionId) and so when they send a request to this lambda, that configuration will be versionId + 1.

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
douglas1850: We need the to_dict and from_dict methods to swap between the data types to store this and then parse it from DDB. Are you seeing a way we can sidestep needing those methods?

lib/stacks/api/appConfiguration.ts
douglas1850: 99% of the time we'll only want one config, but we want to display to Admins a history of prior configs (like the 5 latest) so they can see how the config has been modified over time and optionally perform a rollback if necessary.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
douglas1850: Good question! 

If you put it in the website bucket then it will only get pulled when they do a hard refresh. So if admins are making config changes they're not going to be reflected in the browser for any active users.

You also need the config to be readily available for the backend because when users make requests we need to grab the latest config in the backend to make sure they're not making invalid requests. 

We also mitigate this concern by caching the results from `get_configuration` - most requests won't even make it to Lambda, APIGW will see a request for GET and return the cached response it already has.
* we mitigate returning outdated configs by invalidating/clearing the cache when an update is made

lib/stacks/infra/core.ts
douglas1850: 1. Audting - our customers like being able to audit things and look at the config history, so this way we have a clear trail
2. Rollbacks - if someone pushes a config that does something they don't like it's very easy to rollback to the prior version
3. Metadata - using a DB makes it trivially easy to store metadata for config changes (who changed, the reason why)

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
douglas1850: Good point, I think that could be a nice addition!

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
douglas1850: An upcoming change will be to update/improve our existing admin page such that there's a section for in-app configurations. This will be where this update API is invoked from. It's also where we invoke the get API and request the 5 latest versions so admins can look at prior versions.

In the distant future we'll also create a ProjectOwner page where project owners can see some of these fields and have a project-specific configuration, separate from the app-wide one.

AlejandroRigau: will this admin page be accessible before setting up OIDC, removing the need for the command line setup wizard?

douglas1850: OIDC is necessary to log into the app and is part of configuration at deployment time. The in-app configuration changes will deal with configurations that only affect the app after deployment (such as which services are enabled or disabled). The config setup wizard is still necessary since it deals exclusively with deployment-time configuration.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -0,0 +1,64 @@
...
+
+    try:
+        # Create a record for this configuration and mark it the latest config by incrementing the version Id
+        app_configuration_dao.create(config=app_configuration)
+    except Exception as e:
+        if e.response["Error"]["Code"] == "ConditionalCheckFailedException":
+            e.response["ResponseMetadata"]["HTTPStatusCode"] = 429
+            raise e
+        else:
+            raise e
```
dustins, line 53: If we're doing `raise e` in both scenarios can't we just do this?

```python
if e.response["Error"]["Code"] == "ConditionalCheckFailedException":
    e.response["ResponseMetadata"]["HTTPStatusCode"] = 429

raise e
```

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -0,0 +1,64 @@
...
+
+    return f"Successfully updated configuration for {configScope}, version {version_id}."
+
+
+@api_wrapper
+def get_configuration(event, context):
+    request = json.loads(event["body"])
+    configScope = request["configScope"]
+    num_versions = request.get("numVersions", 1)
```
dustins, line 62: When would they want to get more than one version?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
douglas1850: Good point, will do that

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -0,0 +1,64 @@
...
+
+log = logging.getLogger(__name__)
+app_configuration_dao = AppConfigurationDAO()
+
+
+@api_wrapper
+def update_configuration(event, context):
+    request = json.loads(event["body"])
+    configScope = request["configScope"]
+    version_id = request["versionId"] + 1  # increment so this will be the latest version
```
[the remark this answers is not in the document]
douglas1850: I wanted to increment before we instantiate the AppCofnigurationModel so that the latest version ID is always in the object, as opposed to incrementing it sometime later.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
douglas1850: For the admin page we're going to display several (probably 5) versions so admins can see how the config has changed over time and they can rollback to a prior version if they desire.

dustins (approved):

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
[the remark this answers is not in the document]
dustins: Fair enough, it's just personal preference thing and I wanted to throw out the idea.

dustins (approved):

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
KyleLilly: We probably want to only support multiple versions if the user is an admin otherwise block them from requesting multiples. Not for this PR but for when you actually add this functionality.

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -266,7 +271,13 @@ def lambda_handler(event, context):
                     except Exception as e:
                         logging.exception(e)
                         logging.info("Access Denied. Encountered error while determining resource access policy.")
-
+            elif requested_resource.startswith("/app_configuration"):
```
KyleLilly, line 274: We'll actually need to have this route open to everyone authenticated or not once the OIDC configuration is manageable post deployment. Doesn't need to change in this PR just calling it out as we build out the actual configuration pieces.

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -0,0 +1,387 @@
...
+        application_object_list = []
+        for application in application_list:
+            application_object_list.append(EMRApplications(application["Name"]))
+        return application_object_list
+
+
+class ProjectCreation:
+    def __init__(self, admin_only: bool, allowed_groups: list):
+        self.admin_only = admin_only
+        self.allowed_groups = allowed_groups
```
KyleLilly, line 326: Should this be `allowed_users` or should we add it in addition to what you have here? We don't actually have the concept of groups at the moment.

backend/src/ml_space_lambda/data_access_objects/dynamo_data_store.py
```diff
@@ -63,9 +67,14 @@ def _query(
         index_name: Optional[str] = None,
         limit: Optional[int] = None,
         page_response: bool = False,
+        scan_index_forward: bool = True,
```
KyleLilly, line 70: I did a quick spot check and I think we are always using named args when calling this but generally adding a new optional parameter before an existing parameter can cause issues.

lib/utils/initialAppConfig.ts
```diff
@@ -0,0 +1,101 @@
...
+                'notebook-instance': {'BOOL': 'True'}, 
+                'real-time-translate': {'BOOL': mlspaceConfig.ENABLE_TRANSLATE ? 'True' : 'False'}, 
+                'training-job': {'BOOL': 'True'}, 
+                'batch-translate-job': {'BOOL': mlspaceConfig.ENABLE_TRANSLATE ? 'True' : 'False'}, 
+                'model': {'BOOL': 'True'}, 
+                'hpo-job': {'BOOL': 'True'}, 
+                'endpoint-config': {'BOOL': 'True'}
+            }}, 
+            'ProjectCreation': {'M': {
+                'AdminOnly': {'BOOL': 'True'},
```
KyleLilly, line 25: This should be false right? Currently we allow everyone to create projects by default.

lib/constants.ts
```diff
@@ -25,6 +25,7 @@ export const PROJECT_USERS_TABLE_NAME = 'mlspace-project-users';
 export const USERS_TABLE_NAME = 'mlspace-users';
 export const RESOURCE_SCHEDULE_TABLE_NAME = 'mlspace-resource-schedule';
 export const RESOURCE_METADATA_TABLE_NAME = 'mlspace-resource-metadata';
+export const APP_CONFIGURATION_TABLE_NAME = 'mlspace-app-configuration'
```
KyleLilly, line 28: We need to make sure we document this table in the docs.

backend/test/app_configuration/test_get_configuration.py
```diff
@@ -0,0 +1,58 @@
...
+        ),
+    }
+
+
+def generate_app_config_model(config_scope: str):
+    return AppConfigurationModel(config_scope, 1, None, "tester", "testing", mock_time)
+
+
+@mock.patch("ml_space_lambda.app_configuration.lambda_functions.app_configuration_dao")
+def test_get_config_success(mock_app_config_dao):
```
KyleLilly, line 49: We'll want a failure test case as well.

backend/test/app_configuration/test_update_configuration.py
```diff
@@ -0,0 +1,155 @@
...
+    success_response = f"Successfully updated configuration for {config_scope}, version {version_id+1}."
+    expected_response = generate_html_response(200, success_response)
+
+    assert lambda_handler(mock_event, mock_context) == expected_response
+
+
+@mock.patch("ml_space_lambda.app_configuration.lambda_functions.app_configuration_dao")
+def test_update_config_outdated(mock_app_config_dao):
+    version_id = 1
+    mock_event = generate_event("global", version_id)
```
KyleLilly, line 123: Add a test for project configs since they have the same restriction. I know they work the same but if anything changes we'll want to catch the regression.

backend/test/authorizer/test_authorizer.py
```diff
@@ -1636,6 +1636,54 @@ def test_config_routes(mock_user_dao, user: UserModel, method: str, allow: bool)
...
+            resource="/app_configuration",
+            method=method,
+            path_params=path_params,
+        ),
+        {},
+    ) == policy_response(allow=allow, user=user)
+    mock_user_dao.get.assert_called_with(user.username)
+    #  test/authorizer/test_authorizer.py::test_app_config_routes[user_get_project_config] - AttributeError: 'ProjectUserModel' object has no attribute 'username'
+    #  test/authorizer/test_authorizer.py::test_app_config_routes[user_update_project_config] - AttributeError: 'ProjectUserModel' object has no attribute 'username'
+    #  test/authorizer/test_authorizer.py::test_app_config_routes[project_owner_update_project_config] - AttributeError: 'ProjectUserModel' object has no attribute 'username'
```
KyleLilly, line 1684: Remove these.

backend/test/data_access_objects/test_app_configuration.py
```diff
@@ -0,0 +1,197 @@
...
+        assert len(from_ddb) == 2
+        # The project configs should be returned in descending order based on version ID
+        # PROJECT_RECORD2 has a versionId of 2, so it should come first
+        assert from_ddb[0].to_dict() == self.PROJECT_RECORD2.to_dict()
+        assert from_ddb[1].to_dict() == self.PROJECT_RECORD1.to_dict()
+
+    def test_get_app_config_not_found(self):
+        assert not self.app_config_dao.get("sample-project", 1)
+
+    def test_create_app_config(self):
```
KyleLilly, line 192: We're not testing the case where we get a condition check failure. I know we've got a unit test for that but we should leverage moto to test that too since it's more representative then us mocking.

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
douglas1850: It's supposed to be `allowed_groups` based on the documentation provided; I was told it's not a feature right now, but it's something we plan to have in the future so I wanted to include it for now and just leave it blank and not leverage it in the frontend until it's ready.

lib/utils/initialAppConfig.ts
douglas1850: Good point, will update

backend/src/ml_space_lambda/data_access_objects/dynamo_data_store.py
douglas1850: Good point, I'll bump it to the bottom just in case.

backend/test/data_access_objects/test_app_configuration.py
douglas1850: I see what you mean, I'll add a test where we try to update using an existing versionId, we should expect a ConditionalCheckFailedException

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
KyleLilly: Yeah groups are definitely on the roadmap just wasn't sure if we wanted to support one off non-admin users but I'm fine with leaving it as is for now.

frontend/docs/admin-guide/install.md
```diff
@@ -1061,6 +1061,7 @@ Use the MLSpace Config Wizard by running `npm run config` and select "Advanced C
 | `PROJECTS_TABLE_NAME` |  Dynamo DB table to hold project related metadata | `mlspace-projects` |
 | `PROJECT_USERS_TABLE_NAME` |  Dynamo DB table to hold project membership related metadata. Including permissions and project/user specific IAM role data. | `mlspace-project-users` |
 | `USERS_TABLE_NAME` |  Dynamo DB table to hold user related metadata | `mlspace-users` |
+| APP_CONFIGURATION_TABLE_NAME |  Dynamo DB table to hold dynamic configuration settings. These are settings than can be modified after the app has been deployed. | `mlspace-app-configuration` |
```
KyleLilly, line 1064:
```suggestion
| `APP_CONFIGURATION_TABLE_NAME` |  Dynamo DB table to hold dynamic configuration settings. These are settings than can be modified after the app has been deployed. | `mlspace-app-configuration` |
```

douglas1850: Good catch, fixing

KyleLilly, line 1064: We'll also want [a section for the table here](https://awslabs.github.io/mlspace/admin-guide/getting-started.html#dynamodb) eventually but I we don't need to do that in this PR since this is just foundational.

KyleLilly (dismissed):

lib/stacks/api/appConfiguration.ts
```diff
@@ -0,0 +1,77 @@
...
+                path: 'app-config',
+                method: 'GET',
```
KyleLilly, line 49: This is a GET but you have the lambda parsing a body. Will need to update lambda to use query string parameters.

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -0,0 +1,387 @@
...
+                {"name": size.name, "size": size.size, "masterType": size.master_type, "coreType": size.core_type}
+            )
+        return list_of_dicts
+
+    @staticmethod
+    def extract_cluster_sizes(cluster_size_list: list):
+        # Given a list of cluster sizes, parse them into a list of ClusterSize objects
+        cluster_object_list = []
+        for size in cluster_size_list:
+            cluster_object_list.append(ClusterSize(size["name"], size["size"], size["masterType"], size["coreType"]))
```
KyleLilly, line 243: Just want to make sure we're tracking the need to update the EMR creation lambdas since the format of this is changing.

KyleLilly (approved):

AlejandroRigau (approved):

dustins (approved):

dustins (approved):

AlejandroRigau (approved):

PR #61, opened 2024-04-19
add username normalization in frontend and backend

AlejandroRigau:
*Issue #, if available:* AIML-ADC-8218

*Description of changes:*
- Add username normalization in frontend and backend by removing , . = and spaces. 

This allows usernames with special characters like from a DN. Example: `CN=Jeff Smith,OU=Sales,DC=Fabrikam,DC=COM`

User can still create datasets, project, labeling jobs, etc.
Screenshot 2024-04-19 at 4 23 29 PM

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
   rev: 5.13.2
   hooks:
     - id: isort
-      args: ["--profile", "black", "--check-only", "./backend"]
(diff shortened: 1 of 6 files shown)


AlejandroRigau:
I think I still want to make it so that I use the same regex in the front end and backend

KyleLilly (approved):

douglas1850 (approved):

PR #65, opened 2024-04-26
Reworked DatasetBrowser to reduce rerenders

dustins:
*Issue #, if available:*

*Description of changes:*

I was using `useEffect` more than necessary in the DatasetBrowser component and it was causing a cascade of renders every time the widget was updated. I reworked the component to simplify it and remove unnecessary re-renders.

### DIFF
--- a/frontend/src/modules/dataset/dataset-browser.tsx
+++ b/frontend/src/modules/dataset/dataset-browser.tsx
@@ -69,7 +69,7 @@ export function DatasetBrowser (props: DatasetBrowserProps) {
     }, [dispatch]);
 
     const fetchDatasetResources = useCallback((datasetContext: Partial<DatasetContext>, nextToken?: string, existingItems?: (DatasetResource | IDataset)[]) => {
-        if (manageMode !== DatasetBrowserManageMode.Create) {
+        if (!isCreating) {
             return dispatch(getDatasetContents({datasetContext, projectName, username, nextToken})).then((response) => {
                 if (isFulfilled(response)) {
                     // if nextToken is provided we're appending to existing items
@@ -105,7 +105,7 @@ export function DatasetBrowser (props: DatasetBrowserProps) {
             type: DatasetActionType.State,
             payload: {isLoading: false}
         });
-    }, [dispatch, manageMode, notificationService, projectName, username]);
+    }, [dispatch, isCreating, notificationService, projectName, username]);
 
     const fetchDatasets = useCallback((type: DatasetType) => {
         dispatch(getDatasetsList(projectName)).then((response) => {
@@ -140,7 +140,7 @@ export function DatasetBrowser (props: DatasetBrowserProps) {
             filteringText
         };
 
-        if (manageMode !== DatasetBrowserManageMode.Create) {
+        if (!isCreating) {
             switch (displayMode) {
                 case DatasetBrowserDisplayMode.Resource:
                     if (shouldFetch && datasetContext) {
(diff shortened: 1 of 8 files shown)


frontend/src/modules/dataset/dataset-browser.tsx
```diff
@@ -129,94 +125,70 @@ export function DatasetBrowser (props: DatasetBrowserProps) {
...
-            type: DatasetActionType.Pagination,
-            payload: {
-                pagesCount: Math.max(1, Math.ceil(state.filter.filteredItems.length / pageSize)),
-            }
-        });
-    }, [state.filter.filteredItems, pageSize]);
+    }, [fetchDatasetResources, fetchDatasets, manageMode, state.datasetContext, state.filteringText, state.items, state.nextToken]);
 
-    // if pagination changes make sure currentPageIndex isn't greater than pageCount
+    const [firstLoad, setFirstLoad] = useState(true);
```
KyleLilly, line 172: Do we actually need this? If we only want to run `updateDatasetContext(state.datasetContext || {}, '', false);` once can we just pass an empty array to the `useEffect` instead?

frontend/src/modules/dataset/dataset-browser.tsx
```diff
@@ -129,94 +125,70 @@ export function DatasetBrowser (props: DatasetBrowserProps) {
...
-    }, [state.pagination]);
+    }, [firstLoad, state.datasetContext, updateDatasetContext]);
 
-    // if items or filteringText changes recompute filteredItems
-    useEffect(() => {
+    const filteredItems = useMemo(() => state.items.filter((item) => {
+        return !!item.name?.includes(isCreating ? state.filteringText : resourceForPath(state.filteringText));
+    }), [state.items, state.filteringText, isCreating]);
+
+    const pageSize = 10;
```
KyleLilly, line 184: Are we still supporting page size customization with the dataset tables?

frontend/src/modules/dataset/dataset-browser.tsx
```diff
@@ -236,75 +208,49 @@ export function DatasetBrowser (props: DatasetBrowserProps) {
...
-                        filter: state.filter,
-                    }, setState)} counter={
+                        filteringText: state.filteringText,
+                    }, setState, updateDatasetContext)} counter={
                         state.selectedItems?.length > 0
                             ? `(${state.selectedItems.length}/${state.items.length})`
                             : `(${state.items.length})`
                     }>{displayMode}s</Header>}
-                    empty={state.manageMode === DatasetBrowserManageMode.Create ? EmptyState('No files uploaded') : EmptyState('No Entries exist')}
+                    empty={manageMode === DatasetBrowserManageMode.Create ? EmptyState('No files uploaded') : EmptyState('No Entries exist')}
```
KyleLilly, line 231: Same comment related to `isCreating`.

frontend/src/modules/dataset/dataset-browser.tsx
dustins: You absolutely can do that but you get an error from the linter because you are not listing things you're using. This is a little more verbose but I think the "correct" way to go about this. If your concern is just it is unnecessarily verbose I'm cool swapping that with just disabling the linter for that line.

frontend/src/modules/dataset/dataset-browser.tsx
dustins: We aren't right now, we talked about this a few weeks ago and agreed to just set a reasonable default. I have a branch where I am updating the table component so the DatasetBrowser can use it but I need to rebase that with the changes to Dataset we've since made and the autorefresh stuff that was added recently.

frontend/src/modules/dataset/dataset-browser.tsx
KyleLilly: Yeah I think that's one lint rule that really shouldn't always apply so I'm generally fine if we add a comment as to why we're ignoring it. This is fine too so no big deal.

AlejandroRigau (approved):

douglas1850 (approved):

PR #71, opened 2024-05-01
In app config

douglas1850:
*Description of changes:* This is the groundwork for all future frontend changes for the in-app configuration feature.

## Overview

* Created a reducer class to handle interacting with the backend
* Created a model class so it's easy to interact with the data coming from the backend
* In App.tsx we invoke `getConfiguration` so the app config is store into the `appConfig` state variable and is ready/waiting for us everywhere else in the app
* you'll notice that getConfig leverages `configScope` as a parameter - for right now we're only supporting 'global' (app-wide) scope, but in the future we'll want to support project-level configuration as well, so this is out of considering for that eventuality 
* I've refactored the existing `configuration.tsx` file so that we have clearly delineated sections for dynamic configs (configs we can change after deployment) and deployment configs (things we can't change after deploying). I put each configuration-type into its own component.
* in the `dynamic-configuration.tsx` file you'll notice I left a bunch of empty ExpandableSections with TODOs in them - these will be populated by our team in future upcoming tasks. I also have an input where I'm mutating one of the properties of EMRConfig - this is for testing and to demonstrate how to interact with/update the state for anyone else who works on this in the future.
* I've also removed every instance where we check if translate is enabled via environment variables and replace it with the app config since this is the direction things are going
* Lastly I removed caching from API Gateway - it turns out it's a lot more hassle than it's worth for how little caching is going to improve performance in this case.

Here's what the admin page looks like when dynamic config is part of the configuration.tsx file. My latest commit will remove it from this file so there's risk of a user seeing a page which is a WIP.

localhost_3000_Prod


frontend/src/entities/configuration/configuration-reducer.ts
```diff
@@ -0,0 +1,142 @@
...
+    createSlice,
+    isFulfilled,
+    isPending,
+    PayloadAction,
+} from '@reduxjs/toolkit';
+import axios from '../../shared/util/axios-utils';
+import { defaultConfiguration, IAppConfiguration } from '../../shared/model/app.configuration.model';
+
+const initialState = {
+    loadingAppConfigUpdate: false,
```
KyleLilly, line 28: Name seems a bit misleading - can we rename this `appConfigUpdating` or `appConfigUpdatePending`? There isn't really any `loading` happening with this call.

frontend/src/App.tsx
```diff
@@ -49,6 +50,11 @@ export default function App () {
     // Applies custom theming from public/theming.js
     applyTheme(window.custom_theme);
 
+    const dispatch = useAppDispatch();
+    useEffect(() => {
+        dispatch(getConfiguration({configScope: 'global', clearCache: false}));
```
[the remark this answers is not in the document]
douglas1850: I have a default config which basically disables everything; is there a different way we want to handle not being able to load the config? 

Should we even have a specific page we redirect to if the app config can't load saying to consult with a sys admin or something?

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+export function DynamicConfiguration () {
+    const applicationConfig: IAppConfiguration = useAppSelector(appConfig);
+    const configList: IAppConfiguration[] = useAppSelector(appConfigList);
+
+    const formSchema = z.object({//configuration.EMRConfig.clusterSizes[0].name
+        configuration: z.object({
+            EMRConfig: z.object({
+                clusterSizes: z.object({
+                    name: z.string({ required_error: 'A cluster name must be provided.' })
```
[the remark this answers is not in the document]
douglas1850: I imagined we'd have people flesh this out as we implement each feature, I'm thinking I'll just change it to

```
const formSchema = z.object({
        configuration: z.object({
    });
```

for now

frontend/src/entities/configuration/configuration-reducer.ts
douglas1850: I'm actually just going to remove these altogether as the `formSubmitting` field in the `useValidationReducer` seems to work fine. 

I'll probably need to add a few more formSubmitting fields though to handle all the different loading options (updating the config, importing a config, etc)

frontend/src/entities/configuration/configuration-reducer.ts
```diff
@@ -0,0 +1,142 @@
...
+export const getConfiguration = createAsyncThunk(
+    'app_config/get_configuration',
+    async (props: GetAppConfigurationProps) => {
+        const searchParams = new URLSearchParams();
+        searchParams.set('configScope', props.configScope);
+        const requestUrl = `/app-config?${searchParams.toString()}`;
+        
+        return await axios.get<IAppConfiguration[]>(requestUrl);
+    }
+);
```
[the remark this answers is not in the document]
douglas1850: Nice idea

frontend/src/entities/configuration/configuration-reducer.ts
```diff
@@ -0,0 +1,142 @@
...
+    appConfig: defaultConfiguration,
+};
+
+// Actions
+export const getConfiguration = createAsyncThunk(
+    'app_config/get_configuration',
+    async (props: GetAppConfigurationProps) => {
+        const searchParams = new URLSearchParams();
+        searchParams.set('configScope', props.configScope);
+        const requestUrl = `/app-config?${searchParams.toString()}`;
```
dustins, line 41: FWIW if you're using a variable inside `${...}` it implicitly has toString() called and you don't need to call it explicitly.

frontend/src/entities/configuration/deployment-configuration.tsx
```diff
@@ -0,0 +1,206 @@
...
+import React, { useEffect, useState } from 'react';
+import { getEnvironmentConfig } from './configuration.service';
+import { getBase } from '../../shared/util/breadcrumb-utils';
+import { useParams } from 'react-router-dom';
+import Condition from '../../modules/condition';
+import axios from '../../shared/util/axios-utils';
+import NotificationService from '../../shared/layout/notification/notification.service';
+
+export function DeploymentConfiguration () {
+    const [config, setEnvConfig] = useState(defaultEnvConfig as envConfig);
```
dustins, line 40: Looks like `defaultEnvConfig` is already an `envConfig`. Is there a reason we have to cast it?

frontend/src/App.tsx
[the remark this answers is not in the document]
KyleLilly: I think we probably need to actually display something to users that there was an issue and they should contact a system admin. Disabling everything is fine but really if we can't load the config they're not going to have the correct experience and they may not know why/know that it's not what is expected. If I log and config fails to load and now I can't do anything but there is no explanation I'm going to be confused.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -0,0 +1,123 @@
...
+        formValid: false,
+        formSubmitting: false as boolean,
+    });
+
+    const dispatch = useAppDispatch();
+    useEffect(() => {
+        dispatch(listConfigurations({configScope: 'global', numVersions: 5}));
+    }, [dispatch]);
+
+    const handleSubmit = async () => {
```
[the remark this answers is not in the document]
dustins: I think you should be able to avoid that blurb by using the `isValid` result from `useValidationReducer`. As of recently [it is returned](https://github.com/awslabs/mlspace/blob/43b1253b2b865bcaa73d44c255a9fd550c1d3b23/frontend/src/shared/validation/index.ts#L296-L308) so we can avoid validating forms twice.

frontend/src/App.tsx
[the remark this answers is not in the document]
douglas1850: I've added some logic that will display an error notification if the default config has to be applied when the GET call fails.

I tested this by having the GET function always return a 4XX response and successfully displayed the following notification:

>'Error loading app configuration. Restrictive default policy has been applied in its place. Consult with system admin to resolve issue.',

frontend/src/entities/configuration/dynamic-configuration.tsx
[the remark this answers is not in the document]
douglas1850: I'll try using the `isValid` value, let me know if I use it incorrectly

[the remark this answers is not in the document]
douglas1850: Also added notifications alerting the user to:

1. new config updated successfully
2. new config update failed due to a 429 error (you're operating on an outdated config)
3. new config update failed for an unknown reason, generic error message

douglas1850:
Here's what the admin page looks like when dynamic config is part of the `configuration.tsx` file. My latest commit will remove it from this file so there's risk of a user seeing a page which is a WIP.

localhost_3000_Prod

KyleLilly (approved):
I think we'll probably need to do some additional work on the UX around importing/exporting but the original feedback was addressed and we know the UX for this is a work in progress.

frontend/src/App.tsx
```diff
@@ -49,6 +54,21 @@ export default function App () {
...
 
+    
+    useEffect(() => {
+        dispatch(getConfiguration('global'));
+    }, [dispatch]);
+
+    useEffect(() => {
+        if (configLoadError) {
+            notificationService.generateNotification(
+                'Error loading app configuration. Restrictive default policy has been applied in its place. Consult with system admin to resolve issue.',
```
AlejandroRigau, line 65: where do we actually set the the restrictive configuration specified here: `Restrictive default policy has been applied in its place.`

frontend/src/entities/configuration/configuration-reducer.ts
```diff
@@ -0,0 +1,130 @@
...
+                    failedToLoadConfig: true,
+                };
+            })
+            .addMatcher(isPending(listConfigurations), (state) => {
+                return {
+                    ...state,
+                    failedToLoadConfig: false,
+                };
+            })
+            .addMatcher(isFulfilled(updateConfiguration), (state) => {
```
AlejandroRigau, line 96: do we need these matchers? they seem to just return the state as is

frontend/src/entities/configuration/configuration-reducer.ts
```diff
@@ -0,0 +1,130 @@
...
+            .addMatcher(isPending(updateConfiguration), (state) => {
+                return {
+                    ...state,
+                };
+            });
+            
+    },
+});
+
+// Interfaces
```
AlejandroRigau, line 115: nit: Types instead of interfaces? probably obvious anyways

frontend/src/entities/configuration/configuration-reducer.ts
douglas1850: Right now we don't, I left them in since that seems to be the convention in other reducers and in case we need them in the future but I can remove them.

frontend/src/entities/configuration/configuration-reducer.ts
douglas1850: Good point, will rename

AlejandroRigau (approved):

KyleLilly (approved):

PR #79, opened 2024-05-03
[AIML-ADC-8498] Remove excess background refresh intervals

KyleLilly:
*Issue #, if available:*
AIML-ADC-8498
*Description of changes:*
There was an issue where background refreshes caused a component to be remounted which caused additional intervals to be created (even though they were immediately cleared). Occasionally this would cause things to get out of sync. This PR updates the `useEffect` dependencies to avoid remounting. An update to the way conditionals are handled was also necessary as the old code only worked in the situation where the `useEffect` was invoked on every render. 

Tested this by decreasing the time between refreshes and testing both on an entity table, resource details view without a condition (notebooks) and a details view with a condition (training job) ensuring that once the condition was met there were no background refreshes.


dustins (approved):

frontend/src/shared/util/hooks.ts
```diff
@@ -28,18 +28,25 @@ export function useBackgroundRefresh (action: () => void, deps: readonly unknown
...
             }, (window.env.BACKGROUND_REFRESH_INTERVAL || 60) * 1000);
-            
             return () => {
                 clearInterval(timerId);
             };
         }
-    }, [callbackAction, condition]);
+        // We only want to recreate the interval if the condition changes
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [condition]);
```
AlejandroRigau, line 50: why don't we want to recreate this hook when the function changes? what effect does it have when we have it there?

frontend/src/entities/jobs/transform/detail/transform-detail.tsx
```diff
@@ -75,7 +75,7 @@ function TransformDetail () {
     // Refresh data in the background to keep state fresh
     const isBackgroundRefreshing = useBackgroundRefresh(() => {
         dispatch(describeBatchTransformJob(name));
-    }, [dispatch], (transform.TransformJobStatus !== JobStatus.Failed && transform.TransformJobStatus !== JobStatus.Completed));
+    }, [dispatch, transform.TransformJobStatus], (transform.TransformJobStatus !== JobStatus.Failed && transform.TransformJobStatus !== JobStatus.Completed));
```
AlejandroRigau, line 78: why do we have to pass in the status here? wont it re-render anyways when the condition changes?

frontend/src/shared/util/hooks.ts
dustins: Kyle can say for sure but I think it is "correct" to have callbackAction as a dependency. Likely the lambda being passed in counts as being new every render and we should really be doing use`useBackgroundRefresh(useCallback(...))`

douglas1850 (approved):

frontend/src/shared/util/hooks.ts
KyleLilly: Since the component is remounting the callback was being "changed" which was triggering this `useEffect`. In practice that callback won't actually change (currently).

frontend/src/entities/jobs/transform/detail/transform-detail.tsx
KyleLilly: I was seeing an issue with training jobs where it wasn't triggering an updating and a clearing of the interval. I can run some additional tests though.

PR #83, opened 2024-05-08
App config project creation

douglas1850:
*Description of changes:* The ability for non-admins to create projects will now depend on the status in the app config.

If the property `adminOnly` for project creation is set to `false`, then behavior remains as it currently is (everyone can create a project). 

If the property `adminOnly` is set to true:
* the authorizer will prevent POST requests to /project for non-admins
* the "Create Project" button on the projects page will no longer be visible
* manually navigating to <url>/project/create will now redirect you to the 404 page

This PR also contains:
* unit test updates (ensure users can still create projects when adminOnly is false, users are denied when trying to create projects while adminOnly is true, and admins can still create projects when adminOnly is true)
* in the `initialAppConfig` @AlejandroRigau  brought to my attention that `cluster-sizes` was incorrect and should be `clusterSizes` so that's been fixed as well


backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -283,8 +285,15 @@ def lambda_handler(event, context):
             ) and Permission.ADMIN in user.permissions:
                 policy_statement["Effect"] = "Allow"
             elif requested_resource == "/project" and request_method == "POST":
-                # Anyone can create a project
-                policy_statement["Effect"] = "Allow"
+                # Get the latest app config
+                app_config = AppConfigurationModel.from_dict(
+                    app_configuration_dao.get(configScope="global", num_versions=1)[0]
+                )
```
KyleLilly, line 291: We're going to wind up needing to do this in quite a few places in the authorizer. Can we have a central location to call that we can cache this even if it's just a very short lived? Just something with a 10 second cachetools ttl decorator or something would probably be enough to make a big difference.

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -283,8 +285,15 @@ def lambda_handler(event, context):
...
+                # Check if project creation is admin only; if not, anyone can create a project
+                if (
+                    app_config.configuration.project_creation.admin_only and Permission.ADMIN in user.permissions
```
KyleLilly, line 294: We can do the admin check before fetching the config and save a DDB call.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -145,7 +146,14 @@ export function DynamicConfiguration () {
...
                 </ExpandableSection>
                 <ExpandableSection headerText='Project Creation' variant='default' defaultExpanded>
-                    {<pre>TODO</pre>}
+                    <Toggle
+                        onChange={({ detail }) => {
+                            setFields({ 'configuration.ProjectCreation.isAdminOnly': detail.checked });
+                        }}
+                        checked={state.form.configuration.ProjectCreation?.isAdminOnly || false}
+                    >
+                        Admin Only - when enabled, only admins will be able to create new projects.
```
KyleLilly, line 155: Text seems a bit duplicative. Not sold on this:
```suggestion
                        Admin Only - restrict creation of projects to users with the Admin permission.
```

frontend/src/shared/util/permission-utils.ts
```diff
@@ -34,3 +34,11 @@ export const isAdminOrProjectOwner = (user: IUser, projectPermissions: Permissio
         hasPermission(Permission.ADMIN, user.permissions)
     );
 };
+
+export const projectCreationAdminRequired = (isAdminOnly: boolean, user: IUser) => {
```
KyleLilly, line 38: Again not sold on my suggestion but I think the existing name is a bit confusing especially if we expand things based on additional groups/roles which is on the roadmap.
```suggestion
export const enableProjectCreation = (isAdminOnly: boolean, user: IUser) => {
```

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -145,7 +146,14 @@ export function DynamicConfiguration () {
                     />
                 </ExpandableSection>
                 <ExpandableSection headerText='Project Creation' variant='default' defaultExpanded>
-                    {<pre>TODO</pre>}
+                    <Toggle
```
AlejandroRigau, line 149: we probably need accessibility labels here

AlejandroRigau (changes requested):

frontend/src/entities/configuration/dynamic-configuration.tsx
AlejandroRigau: Seems that these are set automatically. resolving

AlejandroRigau (dismissed):

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: Good idea, will do

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: Your suggestion seems fine, maybe "restrict creation of projects to users with Admin permissions" but same difference

frontend/src/shared/util/permission-utils.ts
douglas1850: Good point, this wouldn't make sense if we're integrating user groups in this function in the future

backend/test/authorizer/test_authorizer.py
```diff
@@ -1923,6 +1985,43 @@ def test_create_project(mock_user_dao):
...
     mock_user_dao.get.assert_called_with(MOCK_USER.username)
+    mock_app_configuration_dao.get.assert_called_with(configScope="global", num_versions=1)
+
+
+@mock.patch.dict("os.environ", TEST_ENV_CONFIG, clear=True)
+@mock.patch("ml_space_lambda.authorizer.lambda_function.app_configuration_dao")
+@mock.patch("ml_space_lambda.authorizer.lambda_function.user_dao")
+def test_create_project_normal_user_admin_only(mock_user_dao, mock_app_configuration_dao):
+    mock_user_dao.get.return_value = MOCK_USER
+    mock_app_configuration_dao.get.return_value = [generate_test_config(admin_only=True)]
```
douglas1850, line 1996: Realized I can just parameterize all of these, working on that now.

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: Good idea, looking into the cachetools decorator.

douglas1850: Took me a minute to figure out how to unit test it properly, but I've got it working, will push on next revision.

AlejandroRigau (dismissed):

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -283,8 +286,15 @@ def lambda_handler(event, context):
...
                 policy_statement["Effect"] = "Allow"
             elif requested_resource == "/project" and request_method == "POST":
-                # Anyone can create a project
-                policy_statement["Effect"] = "Allow"
+                if Permission.ADMIN in user.permissions:
+                    policy_statement["Effect"] = "Allow"
+                else:
+                    # Get the latest app config
+                    app_config = _get_app_config()
+                    logger.info(f"Admin only: {app_config.configuration.project_creation.admin_only}")
```
LightSeekerSC, line 294: This log statement isn't especially clear, but non-blocking.

I think something to the effect of "Project creation restricted to Admins only: <>" more accurately reflects the status. Without looking at the line for this log statement, it would be hard to understand the context seeing this in a log file I think. Also a more verbose log here probably eliminates the need for the comment below.

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -600,3 +610,8 @@ def _get_oidc_props(key_id: str) -> Tuple[Optional[str], Optional[str]]:
         raise ValueError("Missing OIDC configuration parameters.")
 
     return (oidc_keys[key_id], oidc_client_name)
+
+
+@cached(cache=TTLCache(maxsize=1, ttl=10))
+def _get_app_config() -> AppConfigurationModel:
```
LightSeekerSC, line 616: This feels like it should be a global util and not limited to the authorizer. I feel like very soon this will need to be used across a bunch of different lambdas based on all the upcoming tasks we have. Is there a reason this is a local function?

backend/test/authorizer/test_authorizer.py
```diff
@@ -134,6 +134,66 @@ def mock_event(
     }
 
 
+def generate_test_config(config_scope: str = "global", is_project: bool = False, admin_only: bool = False) -> dict:
```
LightSeekerSC, line 137: Again, this feels like something that should be put in a test util as I would expect the config to be used across multiple tests in the near future. I would default it to what we expect to be the most common settings and then have a way to input overrides for the default values. This feels like something that shouldn't be localized to the authorizer.

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: I think I just forgot to remove this after testing, thanks for spotting it

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: That's a good point, I'll move it somewhere common and make it public

backend/test/authorizer/test_authorizer.py
douglas1850: Yeah there's actually already one or two similar ones in other tests, although from looking at other tests I haven't ever seen us use a common utils test-data class, I wonder if there's a reason for that... will look into it.

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: Got an interesting error after moving it to `common_functions`

```
ImportError while loading conftest '/Users/odoug/workspace/MLSpace/mlspace/backend/test/conftest.py'.
test/conftest.py:23: in <module>
    from ml_space_lambda.data_access_objects.dataset import DatasetModel
src/ml_space_lambda/data_access_objects/dataset.py:25: in <module>
    from ml_space_lambda.data_access_objects.dynamo_data_store import DynamoDBObjectStore
src/ml_space_lambda/data_access_objects/dynamo_data_store.py:25: in <module>
    from ml_space_lambda.utils.common_functions import retry_config
src/ml_space_lambda/utils/common_functions.py:26: in <module>
    from ml_space_lambda.data_access_objects.app_configuration import AppConfigurationDAO, AppConfigurationModel
src/ml_space_lambda/data_access_objects/app_configuration.py:24: in <module>
    from ml_space_lambda.data_access_objects.dynamo_data_store import DynamoDBObjectStore
E   ImportError: cannot import name 'DynamoDBObjectStore' from partially initialized module 'ml_space_lambda.data_access_objects.dynamo_data_store' (most likely due to a circular import) (/Users/odoug/workspace/MLSpace/mlspace/backend/src/ml_space_lambda/data_access_objects/dynamo_data_store.py)
```

Looks like being in common functions is causing a circular dependency for some reason.

douglas1850: Resolved it by creating a new `app_config_utils` file without any cross dependencies. Unit tests are running/succeeding again

backend/test/authorizer/test_authorizer.py
douglas1850: It doesn't seem like we can import things from other test classes since they're not in scope, so I'm going to leave this as-is (unless you're aware of a work around, I wasn't able to find us doing anything like this elsewhere)

AlejandroRigau (approved):

LightSeekerSC (approved):

PR #85, opened 2024-05-09
Modifying authorizer so anyone can get the app config

douglas1850:
*Description of changes:* Anyone should be able to GET the app config. This change modifies the authorizer so it will allow anyone to make a GET request on the `app-config` resource.

Tested locally on my machine and I no longer get errors on the login screen about the app config not loading.

### DIFF
--- a/backend/src/ml_space_lambda/authorizer/lambda_function.py
+++ b/backend/src/ml_space_lambda/authorizer/lambda_function.py
@@ -68,7 +68,7 @@ def lambda_handler(event, context):
         f"- Method: {request_method}"
     )
 
-    if requested_resource.startswith("/app-config") and request_method == "GET":
+    if requested_resource == "/app-config" and request_method == "GET":
         # Anyone can get the app config
         policy_statement["Effect"] = "Allow"
 
@@ -275,7 +275,7 @@ def lambda_handler(event, context):
                     except Exception as e:
                         logging.exception(e)
                         logging.info("Access Denied. Encountered error while determining resource access policy.")
-            elif requested_resource.startswith("/app-config") and Permission.ADMIN in user.permissions:
+            elif requested_resource == "/app-config" and request_method == "POST" and Permission.ADMIN in user.permissions:
                 # Operations for app-wide configuration can only be performed by admins
                 policy_statement["Effect"] = "Allow"
             elif requested_resource == "/login" and request_method == "PUT":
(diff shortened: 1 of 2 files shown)


backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -57,6 +57,20 @@ def lambda_handler(event, context):
...
+    path_params = event["pathParameters"]
+    request_method = event["httpMethod"]
+
+    logger.info(
+        f"Determining access for Resource: {requested_resource} "
+        f"- PathParams: {json.dumps(path_params) if path_params else 'N/A'} "
+        f"- Method: {request_method}"
+    )
+
+    if requested_resource.startswith("/app-config") and request_method == "GET":
```
AlejandroRigau, line 70: should this just be 
```suggestion
    if requested_resource == "/app-config" and request_method == "GET":
```

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -269,13 +274,9 @@ def lambda_handler(event, context):
...
                         logging.exception(e)
                         logging.info("Access Denied. Encountered error while determining resource access policy.")
-            elif requested_resource.startswith("/app-config"):
-                # All users can get the app-wide configuration
-                if request_method == "GET":
-                    policy_statement["Effect"] = "Allow"
-                # Any other operation for app-wide configuration can only be performed by admins
-                elif Permission.ADMIN in user.permissions:
-                    policy_statement["Effect"] = "Allow"
+            elif requested_resource.startswith("/app-config") and Permission.ADMIN in user.permissions:
```
AlejandroRigau, line 277: Same here
```suggestion
            elif requested_resource == "/app-config" and Permission.ADMIN in user.permissions:
```

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: I used `startswith` since for a GET request we expect the resource to look something like

```
/app-config?configScope&numVersions
```

but this does raise the question of whether I should just be explicit about only allowing 

```
/app-config?global&1
```

Since that's exactly what the request resource should be for getting the MLSpace-wide app config. I think this would be the correct way forward, I'm just going to do some testing.

douglas1850: Good suggestions, I don't think the `startswith` are necessary here so I'll remove them in both places.

LightSeekerSC (approved):

AlejandroRigau (approved):

PR #87, opened 2024-05-10
[AIML-ADC-8605]Migrate System Banner Settings to Frontend

AlejandroRigau:
*Issue #, if available:* AIML-ADC-8605

*Description of changes:*
- Add System Banner Configuration to Dynamic configuration
- Remove system banner configuration form constants.ts
- remove usage of system banner in backend
- replace window.env settings with system banner configuration

Appearance:
Screenshot 2024-05-10 at 10 48 45 AM


lib/constants.ts
```diff
@@ -119,10 +119,6 @@ export const OIDC_VERIFY_SSL = true;
 export const OIDC_VERIFY_SIGNATURE = true;
 // This defaults to the APIGW url but if you're using custom DNS you should set this to that
 export const OIDC_REDIRECT_URI = '';
-// Optional system banner which will be displayed at the top and the bottom of MLSpace
```
douglas1850, line 122: Let's leave these in for now. If I remember correctly, in the dev sync we discussed keeping these for now and marking them as deprecated. We'll keep them for the next release, then remove them for the release following that.

We're good to keep them remove from `restApi` since we don't need to deploy them, but we should keep them in `configTypes` since we'll want to support them in the magic config still as well. 

What we should do with these values is update `lib/utils/initialAppConfig.ts ` so that the default system banner settings are based on these.

```
'SystemBanner': {'M': {
                'isEnabled': {'BOOL':  mlspaceConfig.SYSTEM_BANNER_TEXT !== ''}, 
                'text': {'S': mlspaceConfig.SYSTEM_BANNER_TEXT}, 
                'textColor': {'S': mlspaceConfig.SYSTEM_BANNER_TEXT_COLOR}, 
                'backgroundColor': {'S': mlspaceConfig.SYSTEM_BANNER_BACKGROUND_COLOR}
            }}, 
```

frontend/src/App.tsx
```diff
@@ -51,7 +54,7 @@ export default function App () {
     return (
         <HashRouter basename={baseHref}>
             <ErrorBoundary>
-                { window.env.SYSTEM_BANNER?.text && <SystemBanner position='TOP' /> }
+                { applicationConfig.configuration.SystemBanner!.isEnabled && <SystemBanner position='TOP' /> }
```
LightSeekerSC, line 57: Is this guaranteed to be defined in the config? If yes, then we shouldn't need the `!`. If no, then we shouldn't add the `!`

From what I see in the shared model code. The ! should not be necessary and should be removed.

frontend/src/App.tsx
```diff
@@ -73,7 +76,7 @@ export default function App () {
                     toolsHide={true}
                     breadcrumbs={<BreadcrumbsProvider/>}
                 />
-                { window.env.SYSTEM_BANNER?.text && <SystemBanner position='BOTTOM' /> }
+                { applicationConfig.configuration.SystemBanner!.isEnabled && <SystemBanner position='BOTTOM' /> }
```
LightSeekerSC, line 79: Same comment as above. Remove the `!` if it is unnecessary, which it appears to be.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -148,7 +150,57 @@ export function DynamicConfiguration () {
                     {<pre>TODO</pre>}
                 </ExpandableSection>
                 <ExpandableSection headerText='System Banner' variant='default' defaultExpanded>
-                    {<pre>TODO</pre>}
+                    <SpaceBetween direction='vertical' size='l'>
+                        <Toggle
+                            onChange={({ detail }) => {
+                                setFields({ 'configuration.SystemBanner.isEnabled': detail.checked });
+                            }}
+                            checked={state.form.configuration.SystemBanner!.isEnabled!}
```
LightSeekerSC, line 158: The `!` should not be necessary here either. Why are they included?

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -148,7 +150,57 @@ export function DynamicConfiguration () {
...
+                        </Toggle>
+                        <FormField
+                            label='Banner Text'
+                        >
+                            <Input
+                                onChange={({ detail }) => {
+                                    setFields({ 'configuration.SystemBanner.text': detail.value });
+                                }}
+                                onBlur={() => touchFields(['configuration.SystemBanner.text'])}
+                                value={state.form.configuration.SystemBanner!.text}
```
LightSeekerSC, line 170: Remove `!` if unnecessary

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -148,7 +150,57 @@ export function DynamicConfiguration () {
...
+                                placeholder='Enter system banner text'
+                                disabled={!state.form.configuration.SystemBanner!.isEnabled}
```
LightSeekerSC, line 172: Remove `!` for SystemBanner if unnecessary for the whole file:

`SystemBanner!` -> `SystemBanner`

frontend/src/modules/system-banner/system-banner.tsx
```diff
@@ -15,21 +15,30 @@
...
+    
     const bannerStyle: React.CSSProperties = {
         width: '100%',
         position: 'fixed',
         zIndex: 4999,
         textAlign: 'center',
         padding: '2px 0px',
-        backgroundColor: window.env.SYSTEM_BANNER.backgroundColor,
-        color: window.env.SYSTEM_BANNER.fontColor,
+        backgroundColor: applicationConfig.configuration.SystemBanner!.backgroundColor,
```
LightSeekerSC, line 40: If unneeded, remove `!` from SystemBanner for the whole file:

`SystemBanner!` -> `SystemBanner`

frontend/src/App.tsx
AlejandroRigau: Seems that it was set to optional in the type. I dont see why it has to be set to optional so I will change it to required. The default configuration just sets it to "False" so it should be fine

douglas1850: It's optional because in the future when we have dynamic config for Projects, neither system banners nor project creation settings will be configurable at the project-level and hence we want to make those optional.

The `!` is fine in my view as the default app-wide config has something for this so it won't be undefined, and whenever the GET call succeeds and gives us the latest app config it will have this field as well.

LightSeekerSC (approved):

douglas1850 (approved):

PR #88, opened 2024-05-10
add folder upload that preserves folder structure

AlejandroRigau:
*Issue #, if available:* AIML-ADC-8347

*Description of changes:*
- add folder upload button that preserves folder structure
- Create UploadButton component to contain duplicate button code

Demo:
Recording 2024-05-10 at 17 08 17

Drag and drop functionality
Recording 2024-05-16 at 16 07 00


AlejandroRigau:
Looking for feedback on how to make the UploadButton component better. I dont like how I need to pass state, setState, etc....

I tried creating the component inside the `DatasetBrowserActions` but then setting isFolderUpload would be set equally on both components (life is state was shared)

AlejandroRigau:
Discussed in standup that we do indeed want the drag and drop feature. Converting this to draft and proceeding to work on it

.eslintrc
```diff
@@ -353,6 +353,9 @@
                     "virtualization",
                     "vpc",
                     "vpcs",
+                    "webkit",
+                    "webkitdirectory",
+                    "webkitRelativePath",
```
dustins, line 358: Is this actually necessary or was this line maybe just flagged before `webkit` was added?

LightSeekerSC: Yeah, both directory, relative, and path should be accepted words. Only webkit should be necessary here

frontend/src/entities/dataset/dataset.actions.tsx
```diff
@@ -161,70 +242,14 @@ export const DatasetBrowserActions = (state: Pick<DatasetBrowserState, 'selected
...
-                    }}
-                    onBlur={() => {
-                        //This remains empty to regain keyboard focus
-                        //to the button after user closes the input window
-                        //for accessibility purposes
-                    }}
-                    multiple
-                />
+                <SpaceBetween size='xs' direction='horizontal'>
+                    <UploadButton state={state} setState={setState} updateDatasetContext={updateDatasetContext} isFileUpload={false}>
```
LightSeekerSC, line 246: Do we really need separate buttons for this? Generally when I upload files there aren't separate buttons. Can we not tell from the selected file context whether this is a folder or not?

AlejandroRigau: if we use pure html, then yes, we need to have separate buttons. We can potentially make it just a single button if I use some kind of library

AlejandroRigau:
I have added drag and drop functionality since last time I opened this for review. Opening for review again.

frontend/src/entities/dataset/dataset-drag-and-drop.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+    // Helps with explicit typing
+    function isFile (entry: FileSystemEntry): entry is FileSystemFileEntry {
+        return entry.isFile;
+    }
+
+    return (
+        <>
+            <div className='dropzone display-none' onDrop={handleDrop} onDragOver={dragOverHandler} onDragEnter={dragEnterHandler} onDragLeave={dragLeaveHandler}/>
+        </>
```
dustins, line 150: I don't think there is any reason to wrap the div in `<>`

dustins: Also would it make more sense to [use a ref](https://react.dev/learn/manipulating-the-dom-with-refs#) instead of doing `document.getElementsByClassName('dropzone')[0]`?

frontend/src/entities/dataset/dataset-drag-and-drop.tsx
```diff
@@ -0,0 +1,152 @@
...
+        event.preventDefault();
+        document.getElementsByClassName('dropzone')[0].classList.add('display-none');
+        fileHandler(await getFilesAsync(event.dataTransfer), manageMode, notificationService, {state, setState, updateDatasetContext, setDisableUpload});
+    }
+
+    function dragEnterHandler (event) {
+        document.getElementsByClassName('dropzone')[0].classList.add('drag-over');
+        if (event.target.classList.contains('dropzone')) {
+            event.target.classList.add('drag-over');
+        }
```
dustins, line 40: If we're only applying the event handler to `div.dropzone` what scenario is this if statement not true?

frontend/src/entities/dataset/dataset.actions.tsx
```diff
@@ -58,15 +60,102 @@ function DatasetActions (props?: any) {
     );
 }
 
-export const DatasetBrowserActions = (state: Pick<DatasetBrowserState, 'selectedItems' | 'items' | 'datasetContext' | 'manageMode' | 'filteringText'>, setState: ReduxDispatch<DatasetBrowserAction> | ReactDispatch<DatasetBrowserAction>, updateDatasetContext: UpdateDatasetContextFunction): React.ReactNode => {
+export async function fileHandler (arrayOfFiles: File[], manageMode: DatasetBrowserManageMode | undefined, notificationService: any, {state, setState, updateDatasetContext, setDisableUpload,}: Pick<UploadButtonProperties, 'state' | 'setState' | 'updateDatasetContext' | 'setDisableUpload'>) {
+    const filesToUpload = arrayOfFiles.map((file: File): DatasetResourceObject => ({
+        bucket: '',
+        type: 'object',
+        key: `${state.datasetContext?.location || ''}${file.webkitRelativePath ? file.webkitRelativePath : file.name}`,
```
dustins, line 67: I think we could clean this up a little by using `${file.webkitRelativePath || file.name}` right?

frontend/src/modules/dataset/dataset-browser.tsx
```diff
@@ -207,7 +207,15 @@ export function DatasetBrowser (props: DatasetBrowserProps) {
                 <Condition condition={header !== undefined}>
                     { header }
                 </Condition>
-                
+
+                <Condition condition={manageMode ? [DatasetBrowserManageMode.Create, DatasetBrowserManageMode.Edit].includes(manageMode) : false}>
+                    <div style={{border: 'dashed 1px', padding: '15px'}}>
```
dustins, line 212: If we're already adding a CSS file we should probably move this styling to there too.

frontend/src/entities/dataset/dataset-drag-and-drop.tsx
AlejandroRigau: This should be removed. This is old code. Thanks for catching it!

frontend/src/entities/dataset/dataset-drag-and-drop.tsx
```diff
@@ -0,0 +1,152 @@
...
+            function readReaderContent (reader: FileSystemDirectoryReader) {
+                reading++;
+
+                reader.readEntries(function (entries) {
+                    reading--;
+                    for (const entry of entries) {
+                        readEntry(entry);
+                    }
+
+                    if (reading === 0) {
```
douglas1850, line 129: What happens if `reading` != 0 by the end? Also what's the `resolve` keyword doing a few lines below, just trying to understand what this `reading` variable is doing and its logic.

frontend/src/entities/dataset/dataset.actions.tsx
```diff
@@ -58,15 +60,102 @@ function DatasetActions (props?: any) {
...
+
+const UploadButton = ({state, setState, updateDatasetContext, isFileUpload, disableUpload, setDisableUpload, children}: UploadButtonProperties): React.ReactNode => {
     const dispatch = useAppDispatch();
     const notificationService = NotificationService(dispatch);
+    const {manageMode} = state;
+    const uploadFile: RefObject<HTMLInputElement> = useRef(null);
+
+    
+
+    return <>
```
douglas1850, line 124: What's the `<>` doing here?

AlejandroRigau: Its a fragment so that I don't have to have a div or any "parent" component wrapping these.

frontend/src/entities/dataset/dataset-drag-and-drop.tsx
AlejandroRigau: When reading !== 0, it means that there are still some files being read asynchronously. This can happen when the function is reading the contents of a directory hierarchy, where it encounters subdirectories and needs to read the files within those subdirectories as well.

dustins (approved):

PR #92, opened 2024-05-13
Transition EMR cluster creation to use AppConfig

douglas1850:
*Description of changes:* Right now EMR uses values from the cluster-config.json file for determining cluster configuration at creation. This PR will transition EMR creation to using app config wherever possible.

TODO: In a followup commit I'm going to remove cluster-config.json altogether since it shouldn't be necessary at all anymore.

### DIFF
--- a/backend/src/ml_space_lambda/emr/lambda_functions.py
+++ b/backend/src/ml_space_lambda/emr/lambda_functions.py
@@ -27,17 +27,15 @@ from ml_space_lambda.data_access_objects.project import ProjectDAO
 from ml_space_lambda.data_access_objects.resource_metadata import ResourceMetadataDAO
 from ml_space_lambda.data_access_objects.resource_scheduler import ResourceSchedulerDAO, ResourceSchedulerModel
 from ml_space_lambda.enums import ResourceType
+from ml_space_lambda.utils.app_config_utils import get_app_config
 from ml_space_lambda.utils.common_functions import api_wrapper, generate_tags, query_resource_metadata, retry_config
 from ml_space_lambda.utils.mlspace_config import get_environment_variables, pull_config_from_s3
 
 logger = logging.getLogger(__name__)
 
-s3 = boto3.client("s3", config=retry_config)
 emr = boto3.client("emr", config=retry_config)
 
 resource_metadata_dao = ResourceMetadataDAO()
-
-cluster_config = {}
 
 resource_scheduler_dao = ResourceSchedulerDAO()
 project_dao = ProjectDAO()
@@ -51,26 +49,11 @@ def create(event, context):
     cluster_name = event_body["clusterName"]
     emr_size = event_body["options"]["emrSize"]
     release_version = event_body["options"]["emrRelease"]
+    app_config = get_app_config()
 
     env_variables = get_environment_variables()
 
     custom_ami_id = event_body["options"]["customAmiId"] if "customAmiId" in event_body["options"] else None
-    applications = []
-
-    global cluster_config
-    if not cluster_config:
-        resp = s3.get_object(Bucket=env_variables["BUCKET"], Key=env_variables["CLUSTER_CONFIG_KEY"])
-        cluster_config = json.loads(resp["Body"].read().decode())
-
-    # get list of applications if provided, otherwise use default from
-    # s3 cluster config
-    if event_body.get("options", {}).get("applications", []):
-        application_list = event_body["options"]["applications"]
-        for app in application_list:
-            app_name = {"Name": app}
-            applications.append(app_name)
-    else:
-        applications = cluster_config["applications"]
 
     # use subnet if provided, otherwise choose a random subnet
     # from s3 config file
@@ -79,20 +62,17 @@ def create(event, context):
         param_file = pull_config_from_s3()
         subnet = random.sample(param_file["pSMSSubnetIds"].split(","), 1)[0]
 
-    # Get Custom EC2 Key Pair
-    ec2_key_name = ""
-    if cluster_config["ec2-key"] != "EC2_KEY":
-        ec2_key_name = cluster_config["ec2-key"]
-
-    master_instance_type = cluster_config[emr_size]["master-type"]
-    core_instance_type = cluster_config[emr_size]["core-type"]
-    instance_count = cluster_config[emr_size]["size"]
+    for size in app_config.configuration.emr_config.cluster_sizes:
+        if size.name == emr_size:
+            master_instance_type = size.master_type
+            core_instance_type = size.core_type
+            instance_count = size.size
 
     args = {}
     args["Name"] = cluster_name
     args["LogUri"] = f"s3://{env_variables['LOG_BUCKET']}"
     args["ReleaseLabel"] = release_version
-    args["Applications"] = applications
+    args["Applications"] = app_config.configuration.emr_config.to_dict()["applications"]
     args["Instances"] = {
         "InstanceGroups": [
             {
@@ -110,8 +90,8 @@ def create(event, context):
                 "InstanceCount": instance_count,
                 "AutoScalingPolicy": {
                     "Constraints": {
-                        "MinCapacity": cluster_config["auto-scaling"]["min-instances"],
-                        "MaxCapacity": cluster_config["auto-scaling"]["max-instances"],
+                        "MinCapacity": app_config.configuration.emr_config.auto_scaling.min_instances,
+                        "MaxCapacity": app_config.configuration.emr_config.auto_scaling.max_instances,
                     },
                     "Rules": [
                         {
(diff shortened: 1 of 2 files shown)


backend/src/ml_space_lambda/emr/lambda_functions.py
```diff
@@ -158,7 +138,7 @@ def create(event, context):
                 },
             },
         ],
-        "Ec2KeyName": ec2_key_name,
+        "Ec2KeyName": "",
```
dustins, line 141: It looks like the `cluster_config` we were reading from s3 before could possibly set this even though it defaulted to empty. Is this an ability to customers we aren't offering anymore?

lib/resources/config/cluster-config.json
dustins, line 1: Do we have any automatic migration path from a customer defined cluster config in JSON to the AppConfig? If not we should provide some a blurb in upgrade notes. As an aside…do we have upgrade notes?

dustins (dismissed):

backend/src/ml_space_lambda/emr/lambda_functions.py
douglas1850: Asked Kyle and he said as far as he's aware nobody was using this. If it's something we still want to support I think we can just move it to a field in `constants.ts` and let it be a deployment time configuration.

lib/resources/config/cluster-config.json
LightSeekerSC: We should have an update path. It is more work, but we have to keep in mind that now we are onboarding a lot of customers on v1.4. Any breaking changes incur negative customer experience even if we provide instructions to fix them. This will be even more compounding if they have to manually migrate all the config changes in this update.

My understanding is that the migration path is that once the default config is created then this should have base values correct? So if they customized this before, then they will need to update their config settings with the dynamic config, but it shouldn't be breaking correct?

douglas1850: I'll look into what it would take to parse the values out of cluster-config.json into the initial app config we deploy... if that's not doable, we'll just add release notes.

backend/src/ml_space_lambda/emr/lambda_functions.py
LightSeekerSC, line 141: Why is this always `""` now? How does using the dynamic config mean this is no longer variable?

lib/resources/config/cluster-config.json
douglas1850: Added some logic so that:

1. if the customer has an existing `cluster-config.json` file we'll parse out their existing values as the initial values for the app config
2. the EC2 SSH key field `Ec2KeyName` will now have a value in `constants.ts` (EMR_EC2_SSH_KEY) for configuring it at deployment time
3. EMR_EC2_SSH_KEY will be set to whatever `ec2-key` is set to in `cluster-config.json` if it exists

backend/src/ml_space_lambda/emr/lambda_functions.py
douglas1850: This will now be defined by the new `EMR_EC2_SSH_KEY` deployment variable. 

When we generate our config object we'll also see if the customer has an existing `cluster-config.json` file, and if they do we'll try to use the `ec2-key` value from that file as the default value for EMR_EC2_SSH_KEY.

dustins (dismissed):

lib/utils/initialAppConfig.ts
```diff
@@ -15,9 +15,35 @@
...
+            fs.readFileSync('lib/resources/config/cluster-config.json').toString('utf8')
+        );
+        if (clusterConfig['applications']) {
+            for (const application of clusterConfig['applications']) {
+                applicationList.push({'M': {'name': {'S': application['Name']}}});
+            }
+        }
+    }
+    // If the cluster config doesn't exist, default to this list
+    if (!clusterConfig) {
```
LightSeekerSC, line 36: This can probably just be in the `else` of the above conditional correct? If the above is true then `clusterConfig` is set.

lib/utils/initialAppConfig.ts
```diff
@@ -15,9 +15,35 @@
...
 
 import { MLSpaceConfig } from './configTypes';
+import * as fs from 'fs';
 
 
 export function generateAppConfig (mlspaceConfig: MLSpaceConfig) {
+    //Check for properties set in config.json and default to that value if it exists
+    let clusterConfig = undefined;
+    let applicationList = [];
+    if (fs.existsSync('lib/resources/config/cluster-config.json')) {
```
LightSeekerSC, line 25: I am still a little new to the app config, so forgive my fuzziness on this. Is this `generateAppConfig` for the initial config? If so, then why would the `cluster-config.json` not exist, since it is included in this release?

backend/src/ml_space_lambda/emr/lambda_functions.py
```diff
@@ -225,6 +225,47 @@ def list_all(event, context):
     return query_resource_metadata(resource_metadata_dao, event, ResourceType.EMR_CLUSTER)
 
 
+@api_wrapper
+def list_applications(event, context):
+    # EMR doesn't seem to have an API for listing the applications available for a specific release label.
+    # If you describe a release label you can get the applications it supports, but we just want to display
```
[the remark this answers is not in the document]
douglas1850: When creating an EMR cluster you choose a release label (ex 6.2.0, 6.10.0, etc)

You can describe a release label to get the list of applications for *that* release label, but there's no way to list all applications (since they vary slightly based on release label).

This is a list of applications that are supported for all 6.X releases (the only ones we're supporting in MLSpace).

backend/src/ml_space_lambda/emr/lambda_functions.py
```diff
@@ -225,6 +225,47 @@ def list_all(event, context):
...
+        "Zeppelin",
+        "Zookeeper",
+    ]
+
+
+@api_wrapper
+def list_release_labels(event, context):
+    response = emr.list_release_labels(
+        Filters={
+            "Prefix": "emr-6",
```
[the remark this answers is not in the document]
douglas1850: If and when version 7 is added to cleared regions we can do this, but based on what Patrick told me we're a ways away from this. We would also need to reevaluate the application list to make sure all applications are still supported across all versions for both release labels.

backend/test/emr/test_list_applications.py
```diff
@@ -0,0 +1,59 @@
...
+    "AWS_DEFAULT_REGION": "us-east-1",
+}
+
+mock_context = mock.Mock()
+
+with mock.patch.dict("os.environ", TEST_ENV_CONFIG, clear=True):
+    from ml_space_lambda.emr import lambda_functions as emr_handler
+
+
+MOCK_APPLICATION_LIST = [
```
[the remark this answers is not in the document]
douglas1850: I like the idea of having the list in some common place, I'll move it so we can re-use it for testing

backend/test/emr/test_list_release_labels.py
```diff
@@ -0,0 +1,65 @@
...
+def test_list_emr_clusters_success(mock_emr):
+    mock_emr.list_release_labels.return_value = {"ReleaseLabels": MOCK_RELEASE_LABEL_LIST}
+    expected_response = generate_html_response(
+        200,
+        {"ReleaseLabels": MOCK_RELEASE_LABEL_LIST},
+    )
+    assert list_release_labels({}, mock_context) == expected_response
+    mock_emr.list_release_labels.assert_called_with(
+        Filters={
+            "Prefix": "emr-6",
```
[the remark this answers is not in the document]
douglas1850: I can move `emr-6` to a common place as well

backend/test/emr/test_list_release_labels.py
```diff
@@ -0,0 +1,65 @@
...
+        "Error": {"Code": "MissingParameter", "Message": "Dummy error message."},
+        "ResponseMetadata": {"HTTPStatusCode": "400"},
+    }
+    operation = "ListReleaseLabels"
+    expected_response = generate_html_response(
+        "400",
+        f"An error occurred (MissingParameter) when calling the {operation} operation: Dummy error message.",
+    )
+    exception_response = ClientError(error_msg, operation)
+    mock_emr.list_release_labels.side_effect = exception_response
```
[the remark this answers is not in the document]
douglas1850: It's the standard client error test we do for all of our backend tests. Probably to make sure the function is leveraging the @apiwrapper

lib/resources/config/cluster-config.json
```diff
@@ -0,0 +1,60 @@
...
+		},
+		"scale-in": {
+			"increment": -1,
+			"percentage-mem-available": 75.0,
+			"eval-periods": 1,
+			"cooldown": 300
+		}
+	},
+	"ami" : "CUSTOM_AMI",
+	"release" : "emr-6.2.0",
```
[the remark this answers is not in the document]
douglas1850: This is actually just the existing cluster-config.json. We don't use this release value, but I'm leaving it in there since we want to keep around `cluster-config.json` for one more release while we migrate customers to using app config.

lib/utils/initialAppConfig.ts
douglas1850: We want to move away from using cluster-config.json while still supporting customers who have cluster-config.json in the meantime. Even once we remove cluster-config.json from our release, there may still be some customers using a cluster-config.json who skipped a release and I want to make sure we're still supporting migration for them.

backend/test/emr/test_list_release_labels.py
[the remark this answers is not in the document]
douglas1850: Looking at it, I don't think there's a good place for it in the backend and I don't think this is something we want in constants.ts right now since it's not actually configurable - anything lower than 6.X would possibly break since we don't support older versions, and 7.X isn't in region yet.

If 7.X is ever built into ADC regions and customers are interested in using it, putting in some work to make this configurable makes sense but for now I think it would be more confusing to make this a configuration that in reality can't be configured.

lib/utils/initialAppConfig.ts
douglas1850: ah yeah good point will do that

dustins (approved):

AlejandroRigau (approved):

PR #94, opened 2024-05-17
Add upload notification while there is an file upload going on

AlejandroRigau:
*Issue #, if available:* AIML-ADC-8952

*Description of changes:*
- Add upload notification while there is an file upload going on. Allows the user to navigate away while the upload is happening. 
- Add border radius to the drop-box to match radius of the containers. 

Demo:
Recording 2024-05-17 at 15 38 50

### DIFF
--- a/frontend/src/entities/dataset/dataset.css
+++ b/frontend/src/entities/dataset/dataset.css
@@ -30,4 +30,5 @@
 .drop-box {
     border: dashed 1px;
     padding: 15px;
+    border-radius: var(--border-radius-container-wqv1zi);
 }
--- a/frontend/src/entities/dataset/dataset.service.tsx
+++ b/frontend/src/entities/dataset/dataset.service.tsx
@@ -19,6 +19,8 @@ import { DatasetType, IDataset } from '../../shared/model/dataset.model';
 import axios from '../../shared/util/axios-utils';
 import { default as Axios } from 'axios';
 import { DatasetContext } from '../../shared/util/dataset-utils';
+import { ProgressBar } from '@cloudscape-design/components';
+import React from 'react';
 
 export const getPresignedUrls = async (data: any) => {
     const requestUrl = '/dataset/presigned-url';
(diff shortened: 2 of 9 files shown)


douglas1850 (dismissed):

AlejandroRigau:
it seems that I have a bug that makes the app crash if I refresh while file is uploading. Need to fix this first before getting approvals.

frontend/src/entities/dataset/dataset.css
```diff
@@ -30,4 +30,5 @@
 .drop-box {
     border: dashed 1px;
     padding: 15px;
+    border-radius: var(--border-radius-container-wqv1zi);
```
dustins, line 33: I don't know much about styling cloudscape styling, how sure are we that this variable is stable?

AlejandroRigau: I spoke with the cloudscape team and it seems I should be using design tokens instead: https://cloudscape.design/foundation/visual-foundation/design-tokens/

LightSeekerSC:
> it seems that I have a bug that makes the app crash if I refresh while file is uploading. Need to fix this first before getting approvals.

Keep in mind that this may just be a hard limitation. If you can't find a way to mitigate this, it is probably fine to put "please keep this page open and do not refresh as your files are uploading"

frontend/src/entities/dataset/dataset.service.tsx
```diff
@@ -144,6 +146,17 @@ export async function uploadResources (datasetContext: DatasetContext, resourceO
 
             if (uploadResponse.status === 204) {
                 successCount++;
+                notificationService.generateNotification(
+                    `Successfully uploaded ${successCount}/${resourceObjects.length} file(s).`,
+                    'in-progress',
+                    'upload-notification',
+                    (<ProgressBar
```
LightSeekerSC, line 153: Very cool! Love this! We may be able to use this in more places around the application.

AlejandroRigau:
This is now ready for review

frontend/src/shared/layout/notification/notification.tsx
```diff
@@ -31,7 +31,7 @@ function NotificationBanner () {
     const dispatch = useDispatch();
     const notificationService = NotificationService(dispatch);
 
-    useEffect(() => {
+    useMemo(() => {
```
dustins, line 34: I actually think `useEffect` is more appropriate since you're not using the output. But I don't think it makes a difference.

dustins (approved):

frontend/src/shared/layout/notification/notification.tsx
AlejandroRigau: I'm using useMemo since it runs on first page load rather than running as a side effect of the page loading. It helps me fix some null errors when passing in components to notifications props.content

douglas1850 (approved):

PR #95, opened 2024-05-19
Emr config

douglas1850:
*Description of changes:* Enable EMR app config on the frontend.

* Admins can now update the EMR Config settings. This includes adding/removing new sizes, modifying the auto-scaling policies, and adding/removing applications.
* The EMR create page will now display sizes based on the app config

See slack for GIF, it was too big for Github.


frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -136,15 +216,225 @@ export function DynamicConfiguration () {
...
-                    {<pre>TODO</pre>}
-                    <Input
-                        data-cy='test-input'
-                        value={state.form.configuration.EMRConfig.clusterSizes[0].name}
-                        onChange={(event) => {
-                            setFields({ 'configuration.EMRConfig.clusterSizes[0].name': event.detail.value });
-                        }}
-                        onBlur={() => touchFields(['configuration.EMRConfig.clusterSizes[0].name'])}
-                    />
+                    <Container
```
KyleLilly, line 219: Looking at the screenshots I think we're overusing containers quite a bit here and we're also missing some whitespace in between things. We really shouldn't be nesting containers when we can help it and the current layout of the config page is quite confusing from a UX perspective (I know some of that was done previously).

douglas1850: I'll start by removing the containers - should I just make more use of the `ExpandableSection` component? Maybe even have those sections collapsed by default so it doesn't take up so much of the screen.

backend/test/emr/test_create_emr_cluster.py
```diff
@@ -98,13 +99,16 @@ def _expected_args(custom_ami: Optional[str] = None):
         "LogUri": "s3://mlspace-log-bucket",
         "ReleaseLabel": MOCK_RELEASE_LABEL,
         "Applications": [{"Name": "Hadoop"}, {"Name": "Spark"}],
+        "ReleaseLabel": MOCK_RELEASE_LABEL,
+        "Applications": [{"Name": "Hadoop"}, {"Name": "Spark"}],
```
dustins, line 103: Are these lines really meant to be repeated? Same for InstanceType below.

douglas1850: I did a merge from main and for some reason it did this in a bunch of places, looks like I missed some. Thanks for catching it!

frontend/src/entities/configuration/cluster-sizes.tsx
```diff
@@ -0,0 +1,164 @@
...
+                    >
+                        Add New Cluster Size
+                    </Button>
+                </SpaceBetween>
+            </Condition>
+        </SpaceBetween>
+    );
+}
+
+type ClusterSizeProps = FormProps<(ClusterSize)>;
```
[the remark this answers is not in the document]
douglas1850: I'm a bit obsessed with parenthesis. I'll remove them

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -205,21 +205,21 @@ def __init__(self, cluster_sizes: list, auto_scaling: EMRAutoScaling, applicatio
...
+            "clusterTypes": ClusterType.to_dict(self.cluster_sizes),
             "autoScaling": self.auto_scaling.to_dict(),
             "applications": EMRApplications.to_dict(self.applications),
         }
 
     @staticmethod
     def from_dict(dict_object: dict) -> EMRConfig:
         return EMRConfig(
-            ClusterSize.extract_cluster_sizes(dict_object["clusterSizes"]),
+            ClusterType.extract_cluster_sizes(dict_object["clusterTypes"]),
```
dustins, line 216: Do we still want to leave it was `extract_cluster_sizes` if we're using the term types now?

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
```diff
@@ -237,10 +237,10 @@ def to_dict(cluster_size_list: list):
 
     @staticmethod
     def extract_cluster_sizes(cluster_size_list: list):
-        # Given a list of cluster sizes, parse them into a list of ClusterSize objects
+        # Given a list of cluster sizes, parse them into a list of ClusterType objects
         cluster_object_list = []
         for size in cluster_size_list:
```
dustins, line 242: Depending on the answer to my other comment this method is still calling stuff `size` a lot that might want to be changed to type for consistency.

frontend/src/entities/configuration/cluster-sizes.tsx
dustins, line 1: Can we change this file name if we're using the term types now?

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -55,13 +86,50 @@ export function DynamicConfiguration () {
...
+            }
+        }
+        setApplicationOptions(appList);
+        // eslint-disable-next-line react-hooks/exhaustive-deps 
+    }, [applicationList, applicationConfig]);
 
     const handleSubmit = async () => {
-        if (isValid) {
+        const parseResult = formSchema.safeParse(state.form);
+        if (parseResult.success) {
```
dustins, line 132: Why change from `isValid` to checking manually? Isn't this doing the exact same thing as is happening in `useValidationReducer`?

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -117,6 +188,15 @@ export function DynamicConfiguration () {
         }
     };
 
+    const addApplication = async (detail: any) => {
```
dustins, line 191: Why is this async?

frontend/src/entities/configuration/cluster-sizes.tsx
```diff
@@ -0,0 +1,163 @@
...
+import React from 'react';
+import { Button, ExpandableSection, FormField, Input, SpaceBetween } from '@cloudscape-design/components';
+import { ClusterType, IAppConfiguration } from '../../shared/model/app.configuration.model';
+import { FormProps } from '../jobs/form-props';
+import { prefixedSetFields, prefixedTouchFields } from '../../shared/validation';
+import Condition from '../../modules/condition';
+import { InstanceTypeSelector } from '../../shared/metadata/instance-type-dropdown';
+
+
+export type ClusterTypesProps = FormProps<(IAppConfiguration)>;
```
dustins, line 25: Why are there parenthesis around this type?

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -136,15 +216,211 @@ export function DynamicConfiguration () {
...
+                    </ExpandableSection>
+                    <ExpandableSection 
+                        headerText='Cluster Types'
+                        variant='default' 
+                        headingTagOverride='h3' 
+                        headerDescription='The cluster types options that users can select from when creating a new Amazon EMR Cluster.'
+                    >
+                        <ClusterTypeConfiguration
+                            item={
+                                state.form as (IAppConfiguration)
```
dustins, line 242: Above if you supply the type you won't have to cast this. If you use `useValidationReducer<IAppConfiguration>(...)` it should know `state.form` is that type. Actually since applicationConfiguration is already that type you can probably just do `form: applicationConfiguration` right?

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -136,15 +216,211 @@ export function DynamicConfiguration () {
...
+                                label='Max Instances'
+                                constraintText='Must be an integer value.'
+                                errorText={formErrors?.configuration?.EMRConfig?.autoScaling?.maxInstances}
+                                description='The maximum number of instances supporting the Amazon EMR Cluster at any time.'
+                            >
+                                <Input
+                                    data-cy='cluster-max-size'
+                                    value={state.form.configuration.EMRConfig.autoScaling.maxInstances.toString()}
+                                    onChange={(event) => {
+                                        setFields({ 'configuration.EMRConfig.autoScaling.maxInstances': +event.detail.value });
```
dustins, line 266: IMHO I think we should use `Number(event.detail.value)` instead so it is a little more explicit/clear what his happening.

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
douglas1850: good point, will change that

backend/src/ml_space_lambda/data_access_objects/app_configuration.py
douglas1850: Yeah looks like there's a lot of places that reference clusterSize, will fix all of those in this file

frontend/src/entities/configuration/cluster-sizes.tsx
douglas1850: Good idea, will do

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: I think I was trying to make use of `formErrors`, I didn't see how we'd be doing that with `isValid` but if you know how we'd do that I'm happy to make the transition

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: Not sure, now that I think of it we'd want this to be synchronous

frontend/src/entities/configuration/cluster-sizes.tsx
douglas1850: I think we established in [this comment](https://github.com/awslabs/mlspace/pull/95#discussion_r1609204972) that I have an unhealthy obsession with parenthesis.

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: Looks like I can just remove the `as (IAppConfiguration)` altogether and it's fine.

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: Yes but then I don't get to show off my Typescript syntax knowledge.

In the interest of readable code I'll make your change here and everywhere else I use the `+` operator.

frontend/src/entities/configuration/dynamic-configuration.tsx
dustins: `useValidationReducer()` returns a [few things](https://github.com/awslabs/mlspace/blob/main/frontend/src/shared/validation/index.ts#L230-L237) including `isValid` and `errors` which I believe is [doing the same thing](https://github.com/awslabs/mlspace/blame/main/frontend/src/shared/validation/index.ts#L296-L303) as your code. So adding back `isValid` and additionally `errors` should give you exactly what.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -23,28 +23,59 @@ import {
...
 import { z } from 'zod';
 import NotificationService from '../../shared/layout/notification/notification.service';
+import { emrApplications, listEMRApplications } from '../emr/emr.reducer';
+import { formatDisplayNumber } from '../../shared/util/form-utils';
+import { ClusterTypeConfiguration } from './cluster-sizes';
 
 export function DynamicConfiguration () {
     const applicationConfig: IAppConfiguration = useAppSelector(appConfig);
     const configList: IAppConfiguration[] = useAppSelector(appConfigList);
+    const applicationList: string[] = useAppSelector(emrApplications);
```
LightSeekerSC, line 42: There is too much name-overlap in the useEffect. I would name this `emrApplications` or `emrApplicationsLIst`

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -55,13 +86,50 @@ export function DynamicConfiguration () {
...
         dispatch(listConfigurations({configScope: 'global', numVersions: 5}));
     }, [dispatch]);
 
+    useEffect(() => {
+        dispatch(listEMRApplications());
+    }, [dispatch]);
+
+    useEffect(() => {
+        const appList: any[] = [];
+        for (const application of applicationList) {
```
LightSeekerSC, line 108: This should probably just be a map: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -55,13 +86,50 @@ export function DynamicConfiguration () {
...
+            const app = {
+                value: application,
+                label: application
+            };
+            appList.push(app);
+        }
+        // Make sure we're always starting from an empty array to prevent duplicates
+        setSelectedApplicationOptions([]);
+        for (const configApp of applicationConfig.configuration.EMRConfig.applications) {
```
LightSeekerSC, line 117: This could also probably be a map: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -55,13 +86,50 @@ export function DynamicConfiguration () {
...
+            if (applicationList.includes(configApp.Name) && !selectedApplicationOptions.includes(configApp.Name)) {
+                const app = {
+                    value: configApp.Name,
+                    label: configApp.Name
+                };
+                setSelectedApplicationOptions((priorSelectedOptions) => [...priorSelectedOptions, app]);
+            }
+        }
+        setApplicationOptions(appList);
```
LightSeekerSC, line 126: Since this isn't modified after line 113, I would move it up to that codeblock for better co-location of code.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -117,6 +188,15 @@ export function DynamicConfiguration () {
         }
     };
 
+    const addApplication = async (detail: any) => {
+        setSelectedApplicationOptions(detail.selectedOptions);
+        const newApps: Application[] = [];
```
LightSeekerSC, line 193: I would probably opt to update this name to `updatedSelectedApps` as "new" implies they are all new which isn't quite accurate here and you can use a [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) here too.

Seems like we do this kind of conversion frequently across the app (I think I remember doing this too). It would probably make sense to create a utility for this that we should be using across the application for `valueToOption` mapper and `optionToValue` mapper

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -215,7 +491,7 @@ export function DynamicConfiguration () {
                     onClick={handleSubmit}
                     loading={state.formSubmitting}
                     data-cy='dynamic-configuration-submit'
-                    disabled={!isValid}
+                    disabled={state.formSubmitting}
```
LightSeekerSC, line 494: Why not both?
`!isValid || state.formSubmitting`

lib/utils/initialAppConfig.ts
```diff
@@ -79,37 +126,37 @@ export function generateAppConfig (mlspaceConfig: MLSpaceConfig) {
...
-                        'name': {'S': clusterConfig['Large'] || 'Large'},
-                        'size': {'N': clusterConfig['Large']['size'] || '7'},
-                        'masterType': {'S': clusterConfig['Large']['master-type'] || 'm5.xlarge'},
-                        'coreType': {'S': clusterConfig['Large']['core-type'] || 'p3.8xlarge'},
+                        'name': {'S': 'Large'},
+                        'size': {'N': String(clusterConfig['Large']['size'])},
+                        'masterType': {'S': clusterConfig['Large']['master-type']},
+                        'coreType': {'S': clusterConfig['Large']['core-type']},
                     }}
                 ]},
```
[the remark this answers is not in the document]
douglas1850: Managed to get something working which will iterate through all of the top level keys (excluding the ones we know about which aren't cluster sizes/types)

frontend/src/entities/configuration/cluster-sizes.tsx
```diff
@@ -0,0 +1,163 @@
...
+            >
+                <InstanceTypeSelector
+                    selectedOption={
+                        {
+                            label: item.masterType,
+                            value: item.masterType,
+                        }
+                    }
+                    onChange={({ detail }) => {
+                        setFields({ masterType: detail.selectedOption.value });
```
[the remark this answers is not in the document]
douglas1850: What kind of validation did you have in mind? 

Eventually we'll update the InstanceTypeSelector component to leverage the instance types allowed in the app config.

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: Not sure how I'd fit all this conditional logic and state change in a map. If you have suggested code I'd be happy to implement it, otherwise this code is functioning and this request is ultimately a preference.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -55,13 +86,50 @@ export function DynamicConfiguration () {
...
+        }
+        // Make sure we're always starting from an empty array to prevent duplicates
+        setSelectedApplicationOptions([]);
+        for (const configApp of applicationConfig.configuration.EMRConfig.applications) {
+            if (applicationList.includes(configApp.Name) && !selectedApplicationOptions.includes(configApp.Name)) {
+                const app = {
+                    value: configApp.Name,
+                    label: configApp.Name
+                };
+                setSelectedApplicationOptions((priorSelectedOptions) => [...priorSelectedOptions, app]);
```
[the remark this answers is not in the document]
douglas1850: Nice idea, will use a temp list

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: Nice idea, create a backlog task for it and we can tackle it across the app once we get app config out the door

frontend/src/entities/configuration/dynamic-configuration.tsx
douglas1850: Good idea, will do both

LightSeekerSC (approved):

PR #96, opened 2024-05-20
fixes issues where datasets were listing incorrectly

dustins:
*Issue #, if available:*

*Description of changes:*

This fixes two separate issues.

- Table (specifically for background refresh) wasn't declaring dependencies properly so changes to `projectName` from `useParams()` wasn't triggering re-computing memoized functions.
- Recent changes to `getDatasetsList` thunk was causing the wrong API endpoint to be hit

### DIFF
--- a/frontend/src/entities/dataset/dataset.reducer.ts
+++ b/frontend/src/entities/dataset/dataset.reducer.ts
@@ -40,9 +40,9 @@ const initialState = {
 
 export const getDatasetsList = createAsyncThunk(
     'dataset/fetch_entity_list',
-    async (params: ServerRequestProps) => {
+    async (params?: ServerRequestProps) => {
         let requestUrl = '/dataset';
-        if (params.projectName !== undefined) {
+        if (params?.projectName !== undefined) {
             requestUrl = `/project/${params.projectName}/datasets`;
         }
         return axios.get<IDataset[]>(requestUrl);
--- a/frontend/src/entities/dataset/dataset.reducer.ts
+++ b/frontend/src/entities/dataset/dataset.reducer.ts
@@ -40,7 +40,7 @@ const initialState = {
 
 export const getDatasetsList = createAsyncThunk(
     'dataset/fetch_entity_list',
-    async (params?: ServerRequestProps) => {
+    async (params: ServerRequestProps) => {
         let requestUrl = '/dataset';
         if (params?.projectName !== undefined) {
             requestUrl = `/project/${params.projectName}/datasets`;
--- a/frontend/src/entities/dataset/dataset.tsx
+++ b/frontend/src/entities/dataset/dataset.tsx
@@ -39,7 +39,6 @@ export const Dataset = () => {
     !projectName ? DocTitle('Datasets') : DocTitle(projectName!.concat(' Datasets'));
 
     useEffect(() => {
-        dispatch(getDatasetsList({projectName}));
         dispatch(
             setBreadcrumbs([
                 getBase(projectName),
--- a/frontend/src/entities/dataset/dataset.tsx
+++ b/frontend/src/entities/dataset/dataset.tsx
@@ -39,7 +39,7 @@ export const Dataset = () => {
     !projectName ? DocTitle('Datasets') : DocTitle(projectName!.concat(' Datasets'));
 
     useEffect(() => {
-        dispatch(getDatasetsList(projectName));
+        dispatch(getDatasetsList({projectName}));
         dispatch(
             setBreadcrumbs([
                 getBase(projectName),
--- a/frontend/src/modules/table/table.types.tsx
+++ b/frontend/src/modules/table/table.types.tsx
@@ -56,7 +56,7 @@ type TableProps<Entry = TableEntry> = {
     focusProps?: any;
     focusFileUploadProps?: any;
     serverRequestProps?: ServerRequestProps;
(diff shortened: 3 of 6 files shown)


frontend/src/modules/table/table.types.tsx
```diff
@@ -56,7 +56,7 @@ type TableProps<Entry = TableEntry> = {
     focusProps?: any;
     focusFileUploadProps?: any;
     serverRequestProps?: ServerRequestProps;
-    serverFetch?: AsyncThunk<any, ServerRequestProps, any>;
+    serverFetch?: AsyncThunk<any, ServerRequestProps | undefined, any>;
```
AlejandroRigau, line 59: this is already optional, do we need to explicitly set it to undefined?

frontend/src/shared/util/hooks.ts
```diff
@@ -53,9 +53,7 @@ export function useBackgroundRefresh (action: () => void, deps: readonly unknown
                 clearInterval(timerId);
             };
         }
-        // We only want to recreate the interval if the condition changes
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [condition]);
+    }, [callbackAction, condition]);
```
AlejandroRigau, line 56: callbackAction was originally left out because it it had the potential to create a lot of setIntervals running at the same time. Maybe I am misremembering but I would like for @KyleLilly to chime in here since he worked on this

frontend/src/shared/util/table-utils.tsx
```diff
@@ -238,7 +238,7 @@ export const ServerSidePaginator = <T extends ServerRequestProps>(props: ServerS
     const loading = props.loading;
     const setLoading = props.setLoading;
 
-    const fetchData = async (usePagination: boolean, callback?: () => void) => {
+    const fetchData = useCallback(async (usePagination: boolean, callback?: () => void) => {
```
AlejandroRigau, line 241: why do we need to memoize this function? seems like it makes no difference to memoize it since its dependent on so many things that would change anyways when we want to call this.

frontend/src/entities/dataset/dataset.reducer.ts
```diff
@@ -40,9 +40,9 @@ const initialState = {
 
 export const getDatasetsList = createAsyncThunk(
     'dataset/fetch_entity_list',
-    async (params: ServerRequestProps) => {
+    async (params?: ServerRequestProps) => {
```
AlejandroRigau, line 43: why do we need to make params optional? it seems we always call `getDatasetsList` with projectName passed in unless I am missing something. Either way, for something like listNotebookInstances we do pass in params exactly like we do before your change. I think we should maintain consistency.

```
export const listNotebookInstances = createAsyncThunk(
    'notebook/list_resources',
    async (params: ServerRequestProps) => {
        let requestUrl = '/notebook';
        if (params.projectName) {
            requestUrl = `/project/${params.projectName}/notebooks`;
        }
        return axios.get<PagedResponsePayload<NotebookResourceMetadata>>(
            addPagingParams(requestUrl, params)
        );
    }
);
```

frontend/src/entities/dataset/dataset.tsx
```diff
@@ -39,7 +39,7 @@ export const Dataset = () => {
     !projectName ? DocTitle('Datasets') : DocTitle(projectName!.concat(' Datasets'));
 
     useEffect(() => {
-        dispatch(getDatasetsList(projectName));
+        dispatch(getDatasetsList({projectName}));
```
AlejandroRigau, line 42: we probably dont need this anymore since we are passing it in the table below

frontend/src/modules/table/table.types.tsx
dustins: Good call, this was a holdover from before and not necessary anymore.

frontend/src/shared/util/hooks.ts
LightSeekerSC: If this is what I remember, then we definitely don't want this back. It can cause a DDOS effect that wrecks the application.

frontend/src/shared/util/table-utils.tsx
dustins: I think you're right we can leave this non memoized.

frontend/src/entities/dataset/dataset.reducer.ts
dustins: This was a holdover from when the parameter was optional. Looks like it doesn't need to be anymore.

frontend/src/entities/dataset/dataset.tsx
dustins: Happy to remove this, I was just fixing a broken function call.

frontend/src/shared/util/hooks.ts
KyleLilly: It was [this](https://github.com/awslabs/mlspace/pull/79/files#r1592849120). I'll pull down this branch and take a quick look though.

dustins: Kyle and I looked at it this morning and while it causes the interval to be recreated way too often that doesn't cause additional network calls or re-renders. Until we have a better solution we're going to let my changes stand as what this fixes is more critical than the potential timing issue it reintroduces.

KyleLilly (approved):

douglas1850 (approved):

PR #97, opened 2024-05-22
Adds App Deny Policy For Disabling Services At An Application Level

LightSeekerSC:
*Description of changes:*
- Adds the `mlspace-app-deny-services-policy` which will be used for controlling whether a service is enabled or disabled via adding and removing deny statements to this policy
- Adds comments that identify what MLSpace service different permissions are associated with
- Refactors to consolidate some redundant or unnecessarily distinct policy statements


lib/stacks/iam.ts
```diff
@@ -385,16 +377,34 @@ export class IAMStack extends Stack {
...
 
+        // The policy that will allow for disabling services by denying IAM permissions at an application level
+        const appDenyServicesPolicy = new ManagedPolicy(this, 'mlspace-app-deny-services-policy', {
+            /**
+             * Policies must have at least one statement with an effect, actions, and resources property for validation
+             * This statement has no effect since the action is invalid
+             */
+            statements: [new PolicyStatement({
+                effect: Effect.DENY,
+                actions: ['x:*'],
```
KyleLilly, line 391: Does this actually deploy as is? Trying to create a policy in the console with `actions: ['x:*']` throws a validation error. Does this need to exist? Can you add some comments here explaining what this is and why it won't cause issues/if it will show up as invalid in the console explain why that's ok?

lib/stacks/iam.ts
```diff
@@ -719,22 +739,14 @@ export class IAMStack extends Stack {
...
-                    new PolicyStatement({
-                        effect: Effect.ALLOW,
-                        actions: ['comprehend:DetectDominantLanguage'],
-                        resources: ['*'],
-                    }),
                 ],
             });
 
-            if (props.enableTranslate) {
-                appPolicy.addStatements(translateIAMPermissionsPolicyStatement);
```
LightSeekerSC, line 734: This is already added in the notebook policy and was duplicative. Consolidated it to save policy space. Verified that translate still works in the UI

lib/stacks/iam.ts
```diff
@@ -822,25 +834,7 @@ export class IAMStack extends Stack {
                         ],
                         conditions: {
                             StringEquals: {
-                                'iam:PassedToService': 'sagemaker.amazonaws.com',
```
LightSeekerSC, line 825: Changing this back. This consolidation works, but allows passing a broader spectrum of roles to translate which is less-scoped. If we need to conserve policy space in the future, this could probably be done.

lib/stacks/iam.ts
```diff
@@ -270,37 +261,38 @@ export class IAMStack extends Stack {
...
+                 * SageMaker permissions to allow users to monitor the status of resources they've
+                 * created. These statements will be supplemented with user/project specific policies
+                 * to ensure users can only describe/interact with resources that have been tagged
+                 * with their username and/or project name.
+                 */
                 new PolicyStatement({
                     effect: Effect.ALLOW,
                     actions: [
-                        // Training Job Actions
+                        // Training Job Service IAM Permissions
```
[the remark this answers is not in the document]
LightSeekerSC: After talking to Dustin the agreed alternative is: `<Service> Permissions`

For permissions related to multiple services the following will be used:
```
/** 
 * <Service 1> Permissions
 * <Service 2> Permissions
 */  
```

LightSeekerSC:
The new desired strategy is not not create this in CDK or have managed policy customers create this manually, but rather this will be dynamically generated on the first update to the Admin app-config, which will create and attach the policy where desired.

estohlmann (approved):

PR #98, opened 2024-05-23
Updating FE code to respect Dynamo config for enabled services

estohlmann:
AIML-ADC-8599

Updating the FE to follow the Dynamo config for enabled/disabled services. This includes services displayed in the side nav, project detail section, and URLs we are allowed to navigate to.

Attached are screenshots with all services enabled, all services disabled, and a demonstration navigating to the blocked Batch Translate URLs. 

All Services Enabled

All Services Disabled

demo

### DIFF
--- a/frontend/src/entities/project/detail/project-detail.tsx
+++ b/frontend/src/entities/project/detail/project-detail.tsx
@@ -62,10 +62,10 @@ function ProjectDetail () {
             .catch(() => {
                 navigate('/404');
             });
-        dispatch(setItemsForProjectName({ 'project': projectName, 'enabledServices': applicationConfig.configuration.EnabledServices }));
+        dispatch(setItemsForProjectName(projectName));
         dispatch(setBreadcrumbs([getBase(projectName)]));
         dispatch(setActiveHref('/#'));
-    }, [dispatch, navigate, projectName, project.name, applicationConfig.configuration.EnabledServices]);
+    }, [dispatch, navigate, projectName, project.name]);
 
     // Refresh data in the background to keep state fresh
     const isBackgroundRefreshing = useBackgroundRefresh(async () => {
--- a/frontend/src/entities/project/detail/project-detail.tsx
+++ b/frontend/src/entities/project/detail/project-detail.tsx
@@ -62,10 +62,10 @@ function ProjectDetail () {
             .catch(() => {
                 navigate('/404');
             });
-        dispatch(setItemsForProjectName(projectName));
+        dispatch(setItemsForProjectName({ 'project': projectName, 'enabledServices': applicationConfig.configuration.EnabledServices }));
         dispatch(setBreadcrumbs([getBase(projectName)]));
         dispatch(setActiveHref('/#'));
-    }, [dispatch, navigate, projectName, project.name]);
+    }, [dispatch, navigate, projectName, project.name, applicationConfig.configuration.EnabledServices]);
 
     // Refresh data in the background to keep state fresh
     const isBackgroundRefreshing = useBackgroundRefresh(async () => {
@@ -209,12 +209,12 @@ function ProjectDetail () {
                                     <ProjectResourceCount iconName='status-stopped' variantStyle='subtle' resourceKey={`${ResourceType.NOTEBOOK}.Stopped`} resourceLabel='Stopped' />
                                 </ProjectResourceRow>
 
-                                <ProjectResourceRow label='EMR clusters' path='emr'>
+                                {applicationConfig.configuration.EnabledServices.emrCluster && <ProjectResourceRow label='EMR clusters' path='emr'>
                                     <ProjectResourceCount iconName='add-plus' variantStyle='subtle' resourceKey={`${ResourceType.EMR_CLUSTER}.Total`} resourceLabel='Total' />
                                     <ProjectResourceCount iconName='status-in-progress' variantStyle='success' resourceKey={`${ResourceType.EMR_CLUSTER}.RUNNING`} resourceLabel='Running' />
                                     <ProjectResourceCount iconName='status-pending' variantStyle='subtle' resourceKey={`${ResourceType.EMR_CLUSTER}.STARTING`} resourceLabel='Starting' />
                                     <ProjectResourceCount iconName='status-positive' variantStyle='success' resourceKey={`${ResourceType.EMR_CLUSTER}.WAITING`} resourceLabel='Waiting' />
-                                </ProjectResourceRow>
+                                </ProjectResourceRow>}
 
                                 <ProjectResourceRow label='Endpoints' path='endpoint'>
                                     <ProjectResourceCount iconName='add-plus' variantStyle='subtle' resourceKey={`${ResourceType.ENDPOINT}.Total`} resourceLabel='Total' />
--- a/frontend/src/entities/routes.tsx
+++ b/frontend/src/entities/routes.tsx
@@ -140,26 +140,34 @@ const EntityRoutes = () => {
                     element={<TransformJobRoutes />}
                 />
 
-                <Route path='project/:projectName/emr' element={<EMRClusters />} />
-                <Route path='project/:projectName/emr/create' element={<EMRClusterCreate />} />
-                <Route
-                    path='project/:projectName/emr/:clusterId/:clusterName'
-                    element={<EMRDetail />}
-                />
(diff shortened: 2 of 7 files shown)


frontend/src/shared/layout/navigation/navigation.reducer.ts
```diff
@@ -51,10 +51,10 @@ const navigationSlice = createSlice({
                 state.adminItems = undefined;
             }
         },
-        setItemsForProjectName (state, action: PayloadAction<string | undefined>) {
-            const projectName = action.payload;
+        setItemsForProjectName (state, action: PayloadAction<any | undefined>) {
```
dustins, line 54: Can we define a type for this so we're not relying on `any`? Looks like `{projectName: string, enabledServices: IEnabledServices}` would probably work.

frontend/src/entities/project/detail/project-detail.tsx
```diff
@@ -62,10 +62,10 @@ function ProjectDetail () {
             .catch(() => {
                 navigate('/404');
             });
-        dispatch(setItemsForProjectName(projectName));
+        dispatch(setItemsForProjectName({ 'project': projectName, 'enabledServices': applicationConfig.configuration.EnabledServices }));
```
LightSeekerSC, line 65: Why are we dispatching this here? Shouldn't this be dispatched when the app config is updated? The reason we have these dispatches here is to add the project name when we end up on a project page.

Additionally, since the reducer should have access to the state, it should be able to access the current appConfig which is part of the store and shouldn't need this passed in unless I am missing someting.

frontend/src/modules/resource-not-found/resource-not-found.tsx
```diff
@@ -22,19 +22,22 @@ import {
...
+import {appConfig} from '../../entities/configuration/configuration-reducer';
 
 export default function ResourceNotFound () {
     const dispatch = useAppDispatch();
+    const applicationConfig: IAppConfiguration = useAppSelector(appConfig);
     DocTitle('Not Found');
 
     useEffect(() => {
-        dispatch(setItemsForProjectName());
+        dispatch(setItemsForProjectName({ 'enabledServices': applicationConfig.configuration.EnabledServices }));
```
LightSeekerSC, line 37: Same comment here as `project-detail`

frontend/src/shared/layout/home/home.tsx
```diff
@@ -25,17 +25,19 @@ import ProjectCards from '../../../entities/project/card';
...
 export const Home = () => {
     const dispatch = useAppDispatch();
-
+    const applicationConfig: IAppConfiguration = useAppSelector(appConfig);
     useEffect(() => {
         dispatch(selectProject());
         dispatch(setBreadcrumbs([]));
         dispatch(setActiveHref('/#'));
-        dispatch(setItemsForProjectName());
+        dispatch(setItemsForProjectName({'enabledServices': applicationConfig.configuration.EnabledServices}));
```
LightSeekerSC, line 38: Same comment as `project-detail`

frontend/src/shared/layout/navigation/navigation.reducer.ts
LightSeekerSC: Do we need the app config value here? The appConfig is in the store, which this reducer should have access to. Why do we need to pass this value in? Additionally if we are directly accessing the current config value, then we don't need to update the input parameters on all of the dispatch calls.

frontend/src/shared/layout/navigation/side-navigation.tsx
```diff
@@ -152,7 +152,7 @@ export default function SideNavigation () {
                         selectedOption={selectedOption}
                         onChange={({ detail }) => {
                             setSelectedOption(detail.selectedOption);
-                            dispatch(setItemsForProjectName(detail.selectedOption.value));
+                            dispatch(setItemsForProjectName({ 'project': detail.selectedOption.value, 'enabledServices': applicationConfig.configuration.EnabledServices }));
```
LightSeekerSC, line 155: Same comment as `project-detail`

frontend/src/entities/routes.tsx
```diff
@@ -140,26 +140,34 @@ const EntityRoutes = () => {
...
                 ) : undefined}
-                <Route
-                    path='project/:projectName/batch-translate/create'
-                    element={<BatchTranslateCreate />}
-                />
-                <Route
-                    path='project/:projectName/batch-translate/:jobId'
-                    element={<BatchTranslateDetail />}
-                />
                 <Route path='*' element={<ResourceNotFound />} />
```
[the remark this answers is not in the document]
estohlmann: Correct I had to make changes here for Batch Translation and EMR. Ground Truth(lines 126-131) and RT Translation(lines 84-86) were already complete.

frontend/src/shared/layout/navigation/navigation.reducer.ts
estohlmann: This is my first experience working with Thunk but I was under the impression that wasn't how it worked. 

> Sometimes a slice reducer really needs to read a value that isn't in its own slice of state. A possible workaround to that is to dispatch a thunk, extract the needed values from state, and then dispatch a plain action containing the additional info.

frontend/src/shared/layout/navigation/side-navigation.tsx
dustins: Especially now that this is a typed object maybe we shouldn't wrap the object properties in strings for all these calls? Feels unnecessary. Probably not worth changing unless you're making another PR though.

frontend/src/shared/layout/navigation/navigation.reducer.ts
```diff
@@ -51,10 +58,10 @@ const navigationSlice = createSlice({
                 state.adminItems = undefined;
             }
         },
-        setItemsForProjectName (state, action: PayloadAction<string | undefined>) {
-            const projectName = action.payload;
+        setItemsForProjectName (state, action: PayloadAction<NavigationReducerState | undefined>) {
```
dustins, line 61: Is this action used anywhere else in the code now where it isn't sent a `NavigationReducerState`? If not we can get rid of the `| undefined`. Also assuming we can do that we can remove all the the `action.payload?.enabledServices?.<blah>` with just `action.payload.enabledServices.<blah>` since it will always be valid.

frontend/src/shared/layout/navigation/navigation.reducer.ts
dustins: @estohlmann As far as I can tell you're not using thunk just a vanilla react reducer which makes sense since you're not making an async call.

@LightSeekerSC I don't believe he will have access to the app config in this reducer since `state` is available but will only be the navigation state/slice and not the top level application state which he would need.

frontend/src/shared/layout/navigation/navigation.reducer.ts
estohlmann: Updated to remove optional undefined param.

dustins (dismissed):

frontend/src/shared/layout/navigation/navigation.reducer.ts
LightSeekerSC: Okay, this makes me not like Redux. @dustins is right though and as the comment @estohlmann quoted, Thunk does allow this. The navigation reducer doesn't use Thunk, but the app config does. I tested this, and it *should* work.

You can import `useAppDispatch` in `configuration-reducer.ts` and then within the `listConfigurations` function, if you change it so that it doesn't immediately return the axios response, but saves it, then you can dispatch the `data[0]` element's `EnabledServices` of the response to update EnabledServices within the navigation element instead of doing it in multiple other places. This is still kind of gross, but it consolidates the logic and I believe that it will work

frontend/src/modules/resource-not-found/resource-not-found.tsx
```diff
@@ -22,7 +22,7 @@ import {
     setBreadcrumbs,
     setItemsForProjectName,
 } from '../../shared/layout/navigation/navigation.reducer';
-import { useAppDispatch } from '../../config/store';
+import {useAppDispatch} from '../../config/store';
```
LightSeekerSC, line 25: I think this was changed on accident, but I won't block if it somehow sneaks by eslint. Seeing this in several imports. I thought that eslint did check for this though.

LightSeekerSC (approved):

dustins (approved):

PR #99, opened 2024-05-23
Updated instance types to allowlist instead of blocklist

dustins:
*Issue #, if available:*

*Description of changes:*
After our discussion this morning we decided we wanted to use an allow list instead of a block list. Updating the model to reflect EnabledInstanceTypes instead of DiabledInstancedTypes now makes more sense.

### DIFF
--- a/backend/test/data_access_objects/test_app_configuration.py
+++ b/backend/test/data_access_objects/test_app_configuration.py
@@ -62,7 +62,7 @@ def generate_test_config(config_scope: str, version_id: int, is_project: bool) -
         "changedBy": "Tester",
         "createdAt": mock_time,
         "configuration": {
-            "DisabledInstanceTypes": {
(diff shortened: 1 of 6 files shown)


douglas1850 (approved):

backend/test/data_access_objects/test_app_configuration.py
```diff
@@ -62,7 +62,7 @@ def generate_test_config(config_scope: str, version_id: int, is_project: bool) -
         "changedBy": "Tester",
         "createdAt": mock_time,
         "configuration": {
-            "DisabledInstanceTypes": {
+            "EnabledInstanceTypes": {
```
LightSeekerSC, line 65: Are we not testing this yet? Confusing how this passes with no other changes. I would expect conditionals to need to be inverted.

LightSeekerSC (approved):

PR #103, opened 2024-05-28
backend changes for constraining instance types

dustins:
*Issue #, if available:*

*Description of changes:*
- added 3 new policies to the notebook role to constraint instance type for the following services
  - 1 for jobs (transform, training, hpo)
  - 1 for notebooks
  - 1 for endpoint configurations
- updated the `update_instance_constraint_policies` to automatically update those policies


lib/stacks/iam.ts
```diff
@@ -383,10 +389,45 @@ export class IAMStack extends Stack {
...
+                    },
+                }),
+            ];
+        };
+
         const notebookPolicy = new ManagedPolicy(this, 'mlspace-notebook-policy', {
             statements: notebookPolicyStatements(this.partition, Aws.REGION),
         });
 
+        this.mlspaceEndpointConfigInstanceConstraintPolicy = new ManagedPolicy(this, 'mlspace-endpoint-config-instance-constraint', {
```
LightSeekerSC, line 415: Same comment here as on the `instanceConstraintPolicyStatement` functyion

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -53,9 +73,88 @@ def update_configuration(event, context):
...
+    return f"arn:{components[1]}:sagemaker:{components[3]}:{components[4]}:{resource}/*"
+
+
+def update_instance_constraint_policies(previous_configuration, new_configuration, context) -> None:
+    env_vars = get_environment_variables()
+
+    if (
+        previous_configuration.training_job_instance_types != new_configuration.training_job_instance_types
+        or previous_configuration.transform_jobs_instance_types != new_configuration.transform_jobs_instance_types
+    ):
```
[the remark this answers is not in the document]
dustins: This is actually getting removed now that we're going to stop trying to detect changes and just update the policy.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -53,9 +73,88 @@ def update_configuration(event, context):
...
 
-@api_wrapper
-def get_configuration(event, context):
-    configScope = event["queryStringParameters"]["configScope"]
-    num_versions = int(event["queryStringParameters"].get("numVersions", 1))
+def create_instance_constraint_statement(
+    statement_id: str, actions: list[str], resources: list[str], allowed_instances: list[str]
+):
+    return {
+        "Sid": statement_id,
```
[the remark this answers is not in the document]
dustins: We're not at any risk of character limits but you're right it doesn't add much. Let the `Sid` be stricken from the record!

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -53,9 +73,88 @@ def update_configuration(event, context):
...
+                "Statement": statements,
+            }
+        ),
+        SetAsDefault=True,
+    )
+
+    delete_non_default_policy(policy_arn)
+
+
+def delete_non_default_policy(policy_arn: str) -> None:
```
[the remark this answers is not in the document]
dustins: Looks like `IAMManager` is already doing essentially the same thing. I can just reuse that I believe.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -33,6 +46,13 @@ def update_configuration(event, context):
 
     new_configuration = SettingsModel.from_dict(request["configuration"])
 
+    previous_app_configurations = app_configuration_dao.get(configScope=configScope, num_versions=1)
```
douglas1850, line 49: Posting here for posterity; Dustin and I discussed doing that same thing as the enabling/disabling services feature is going to do, which is we're not going to bother comparing this updating config against prior configs, but rather we'll simply generate an IAM policy based on whatever the latest config is and update the IAM policy using that.

backend/src/ml_space_lambda/utils/mlspace_config.py
```diff
@@ -43,28 +43,31 @@ def get_environment_variables() -> dict:
...
+            "PROJECT_USERS_TABLE": os.getenv("PROJECT_USERS_TABLE", "mlspace-project-users"),
+            "PROJECTS_TABLE": os.getenv("PROJECTS_TABLE", "mlspace-projects"),
+            "RESOURCE_METADATA_TABLE": os.getenv("RESOURCE_METADATA_TABLE", "mlspace-resource-metadata"),
+            "RESOURCE_SCHEDULE_TABLE": os.getenv("RESOURCE_SCHEDULE_TABLE", "mlspace-resource-schedule"),
+            "S3_KEY": os.getenv("S3_KEY", "notebook-params.json"),
+            "SYSTEM_TAG": os.getenv("SYSTEM_TAG", "MLSpace"),
             "TRANSLATE_DATE_ROLE_ARN": os.getenv("TRANSLATE_DATE_ROLE_ARN", ""),
+            "USERS_TABLE": os.getenv("USERS_TABLE", "mlspace-users"),
+            "ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN": os.getenv("ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN", ""),
+            "NOTEBOOK_INSTANCE_CONSTRAINT_POLICY_ARN": os.getenv("NOTEBOOK_INSTANCE_CONSTRAINT_POLICY_ARN", ""),
```
LightSeekerSC, line 69: If you are attaching this to the notebook role then you don't need this as the dynamic roles will automatically pass any policies attached to that role to the dynamic roles.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -53,9 +73,88 @@ def update_configuration(event, context):
...
-    num_versions = int(event["queryStringParameters"].get("numVersions", 1))
+def create_instance_constraint_statement(
+    statement_id: str, actions: list[str], resources: list[str], allowed_instances: list[str]
+):
+    return {
+        "Sid": statement_id,
+        "Effect": "Allow",
+        "Action": actions,
+        "Resource": resources,
+        "Condition": {"ForAnyValue:StringEquals": {"sagemaker:InstanceTypes": allowed_instances}},
```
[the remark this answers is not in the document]
dustins: Good callout. I think I'm going to leave it as-is now to not add potentially unnecessary complexity. It will be an easy change in the future if we have to refactor due to changing requirements.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -53,9 +73,88 @@ def update_configuration(event, context):
...
+    }
 
-    return app_configuration_dao.get(configScope=configScope, num_versions=num_versions)
+
+def create_sagemaker_resource_arn(resource: str, context) -> str:
```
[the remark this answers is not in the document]
dustins: I thought about it but I think it is pretty specific. I think if anything in the future we'd want an ARN object for easier manipulation (if that doesn't exist already). I'll add backlog item to investigate if we're doing enough manipulation like this to warrant standardizing on something like that.

backend/src/ml_space_lambda/utils/mlspace_config.py
dustins: Yeah this is leftover from when I was constraining the size of notebook instances you can create from notebooks. Since we don't allow this I removed it but forgot to remove this.

lib/stacks/iam.ts
```diff
@@ -383,10 +389,45 @@ export class IAMStack extends Stack {
             return statements;
         };
 
+        const instanceConstraintPolicyStatement = (partition: string, region: string, action: string, resource: string) => {
```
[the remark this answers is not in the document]
dustins: I'll add a comment as a warning. Is there anything we can do to prevent that?

lib/stacks/iam.ts
```diff
@@ -405,7 +446,12 @@ export class IAMStack extends Stack {
             this.mlSpaceNotebookRole = new Role(this, 'mlspace-notebook-role', {
                 roleName: mlSpaceNotebookRoleName,
                 assumedBy: notebookPolicyAllowPrinciples,
-                managedPolicies: [notebookPolicy],
+                managedPolicies: [
+                    notebookPolicy,
+                    this.mlspaceEndpointConfigInstanceConstraintPolicy,
+                    this.mlspaceNotebookInstanceConstraintPolicy,
```
dustins, line 452: This shouldn't be here and will be removed.

lib/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "AWS_ACCOUNT": "427935540279",
+  "AWS_REGION": "us-east-1",
+  "OIDC_URL": "https://cognito-idp.us-east-1.amazonaws.com/us-east-1_gGfd5y9Fd",
```
LightSeekerSC, line 4: This file should probably be removed.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
dustins: Yeah that doesn't need to exist anymore now that we're not comparing.

lib/config.json
KyleLilly: This _should_ be in `.gitignore` - I could have sworn it was but doesn't appear to be in the committed code.

PR #105, opened 2024-05-30
Populate enabled instance types for app config at deployment

douglas1850:
*Description of changes:* since we've swapped the DisabledInstanceTypes list for an EnabledInstanceTypes list, we need to pre-populate the enabled instance types. 

These changes include:
* creating a new lambda function which will get the compute types for each service, put them into the EnabledInstanceType lists for their correlating service, then upload that list to mlspace-app-configuration table
* update the CDK code so that this lambda is executed at creation time

I tested these changes by creating an entry in my mlspace-app-configuration table for a global config of versionID = 0, then deployed the CDK changes. Since the custom CDK resource was being created for the first time it executed and it populated the config so the EnabledInstanceTypes list was fully populated. 

### DIFF
--- a/backend/src/ml_space_lambda/initial_app_config/lambda_function.py
+++ b/backend/src/ml_space_lambda/initial_app_config/lambda_function.py
@@ -1,0 +1,72 @@
+#   Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+#   Licensed under the Apache License, Version 2.0 (the "License").
+#   You may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+# TODO: create a lambda handler. We accept the service type as an argument. We then map that service type to
+# the instance type it needs when we call compute types. We parse out the instances from compute types,
+# put those in a big list, and then use those to make a PUT call to the app config table
+import json
+import logging
+
+import boto3
+from dynamodb_json import json_util as dynamodb_json
+
+from ml_space_lambda.data_access_objects.dynamo_data_store import DynamoDBObjectStore
+from ml_space_lambda.enums import ServiceType
+from ml_space_lambda.metadata.lambda_functions import get_compute_types
+from ml_space_lambda.utils.mlspace_config import get_environment_variables, retry_config
+
+log = logging.getLogger(__name__)
+ddb = boto3.client("dynamodb", config=retry_config)
+
+
+def lambda_handler(event, context):
+    instances = get_compute_types()
+    initial_config_dao = InitialConfigDAO()
+    resp = initial_config_dao.get("global", 1)
+    config = resp[0]["configuration"]
+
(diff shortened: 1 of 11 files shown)


backend/src/ml_space_lambda/initial_app_config/lambda_function.py
```diff
@@ -0,0 +1,69 @@
...
+
+    config["EnabledInstanceTypes"][ServiceType.NOTEBOOK.value] = instances["InstanceTypes"]["InstanceType"]
+    config["EnabledInstanceTypes"][ServiceType.TRAINING_JOB.value] = instances["InstanceTypes"]["TrainingInstanceType"]
+    config["EnabledInstanceTypes"][ServiceType.TRANSFORM_JOB.value] = instances["InstanceTypes"]["TransformInstanceType"]
+    config["EnabledInstanceTypes"][ServiceType.ENDPOINT.value] = instances["InstanceTypes"]["ProductionVariantInstanceType"]
+
+    initial_config_dao.update(config)
+
+
+class InitialConfigDAO(DynamoDBObjectStore):
```
douglas1850, line 44: Realizing there's no reason this can't be folded into the AppConfigDAO, so making that change now.

backend/src/ml_space_lambda/metadata/lambda_functions.py
```diff
@@ -58,6 +58,10 @@
 
 @api_wrapper
 def compute_types(event, context):
+    return get_compute_types()
+
+
+def get_compute_types():
     response = {"InstanceTypes": {}, "AcceleratorTypes": {}}
 
     if this.cached_response_compute_types:
```
LightSeekerSC, line 67: Why are these two separate functions?

lib/stacks/infra/core.ts
```diff
@@ -435,6 +454,31 @@ export class CoreStack extends Stack {
...
+            onCreate: {
+                service: 'Lambda',
+                action: 'invoke',
+                physicalResourceId: PhysicalResourceId.of('initInstanceTypes'),
+                parameters: {
+                    FunctionName: appConfigLambda.functionName,
+                    Payload: '{}'
+                }, 
+            },
+            policy: AwsCustomResourcePolicy.fromStatements([
```
LightSeekerSC, line 467: Like discussed in the Dev Sync, this will fail on customer deployments because their roles aren't allowed to create IAM resources. I think that the App Role would work for this and that is already created.

LightSeekerSC (changes requested):

backend/src/ml_space_lambda/metadata/lambda_functions.py
douglas1850: I wanted to leave the lambda function as-is to not break any APIs, but I also now need to be able to get the compute types without dealing with the API wrapper

lib/stacks/infra/core.ts
douglas1850: okay good idea, will try giving it the app role

backend/src/ml_space_lambda/metadata/lambda_functions.py
AlejandroRigau: can you add a comment to explain that?

AlejandroRigau (changes requested):

backend/src/ml_space_lambda/metadata/lambda_functions.py
douglas1850: What kind of comment are you thinking? I don't like leaving comments in code like "we abstracted to a function because this logic is reused" since code comments are typically to explain what's happening, not why something was done.

estohlmann (approved):

PR #110, opened 2024-05-31
Activate/Deactivate services helper function

LightSeekerSC:
*Description of changes:*
- Adds a helper function for creating the service deactivate policy
- Adds a utility for creating/updating a policy
- Some refactors to the project policy and user policy creation


backend/src/ml_space_lambda/enums.py
```diff
@@ -54,6 +54,43 @@ class ServiceType(str, Enum):
...
+    RESOURCE = "Resource"
+    CONDITION = "Condition"
+
+
+class IAMEffect(str, Enum):
+    DENY = "Deny"
+    ALLOW = "Allow"
+
+
+class EnvVariable(str, Enum):
```
KyleLilly, line 69: If you're going to add this you should [update this](https://github.com/awslabs/mlspace/blob/main/backend/src/ml_space_lambda/utils/mlspace_config.py) at a minimum.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -19,10 +19,107 @@
...
 from ml_space_lambda.utils.common_functions import api_wrapper
+from ml_space_lambda.utils.iam_manager import IAM_RESOURCE_PREFIX, IAMManager
+from ml_space_lambda.utils.mlspace_config import get_environment_variables
 
+iam_manager = IAMManager()
 log = logging.getLogger(__name__)
 app_configuration_dao = AppConfigurationDAO()
+env_variables = get_environment_variables()
+
+group_permissions = [
```
KyleLilly, line 33: Why is this called `group_permissions`?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -50,6 +147,41 @@ def update_configuration(event, context):
             e.response["ResponseMetadata"]["HTTPStatusCode"] = 429
         raise e
 
+    # Create the new app deny policy
+    try:
+        deny_policy_statements = []
+        enabled_services_dict = app_configuration.configuration.enabled_services.to_dict()
+        # Check each service
+        for service in enabled_services_dict.keys():
```
KyleLilly, line 155: If there are no disabled services just bail here rather than trying to create an empty policy.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -50,6 +147,41 @@ def update_configuration(event, context):
...
+
+        iam_manager.create_or_update_policy(
+            iam_manager.generate_policy_string(deny_policy_statements),
+            "app-denied-services",
+            "services",
+            "deny",
+            on_create_attach_to_notebook_role=True,
+        )
+    except Exception as e:
+        logging.error(f"Failed to create or update denied services policy {str(e)}")
```
KyleLilly, line 183: In what scenario would a failure here be acceptable? If we try to deny services and we fail to create the policy/attach it we need to communicate that to users. logging an error and then returning a success message means we'll have now way of letting users know that there was an issue.

backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -41,6 +41,14 @@ def __init__(self, iam_client=None, sts_client=None):
         self.aws_partition = boto3.Session().get_partition_for_region(boto3.Session().region_name)
         self.iam_client = iam_client if iam_client else boto3.client("iam", config=retry_config)
 
+        env_variables = get_environment_variables()
+        self.data_bucket = env_variables["DATA_BUCKET"]
+        self.system_tag = env_variables["SYSTEM_TAG"]
+        self.notebook_role_name = os.getenv("NOTEBOOK_ROLE_NAME", "")
+        self.default_notebook_role_policy_arns = []
+        self.permissions_boundary_arn = os.getenv("PERMISSIONS_BOUNDARY_ARN", "")
+        self.aws_account = self.sts_client.get_caller_identity()["Account"]
```
KyleLilly, line 50: Is there a reason you're not using the `account_utils` you added for this? Same for the enums.

backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -208,21 +218,20 @@ def add_iam_role(self, project_name: str, username: str) -> str:
         project_policy_name = f"{IAM_RESOURCE_PREFIX}-project-{project_name}"
         user_policy_name = f"{IAM_RESOURCE_PREFIX}-user-{username}"
 
-        aws_account = self.sts_client.get_caller_identity()["Account"]
-        project_policy_arn = f"arn:{self.aws_partition}:iam::{aws_account}:policy/{project_policy_name}"
-        user_policy_arn = f"arn:{self.aws_partition}:iam::{aws_account}:policy/{user_policy_name}"
+        project_policy_arn = f"arn:{self.aws_partition}:iam::{self.aws_account}:policy/{project_policy_name}"
```
KyleLilly, line 221: Similar comment re: your new account utils.

backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -435,21 +444,73 @@ def _detach_iam_policies(self, iam_role_name: str) -> List[str]:
...
-            .replace("$USER_NAME", user)
-            .replace("$BUCKET_NAME", self.data_bucket)
-            .replace("$PARTITION", self.aws_partition)
-        )
+    Inputs:
+    policy_template: str - A template that includes variables of the form $VAR_NAME
+    replacments: list[list[str]] - A list of str tuples that are to be replaced of the form ["$VAR_NAME", "var value"]
+    """
+
+    def generate_policy_from_template_string(self, policy_template: str, replacements: list[list[str]]):
```
KyleLilly, line 455: Do all of these need to be called externally? If not I'd recommend adding the `_` prefix back.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: These are groups with shared IAM permissions. These permissions cannot be deactivated unless all services in the group are deactivated. Is there a preferred alternate name for this?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: ACK: This is a draft. Working on figuring out what to do here. The config update should still be successful, so this is a mixed result. Looking to see if @api_wrapper has options for this.

backend/src/ml_space_lambda/utils/iam_manager.py
LightSeekerSC: Did this before creating the `account_utils`. Good callout

backend/src/ml_space_lambda/utils/iam_manager.py
LightSeekerSC: This was a refactor that was done when I thought I would use templates. This could be reverted if we want, but I will keep it and add the `_` for now

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: Updated to `service_group_disable_permissions` so that it more closely resembles the service permissions object naming convention that follows

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: This is now covered. In the case that there are no deactivated services a filler statement will be applied for scalability reasons.

In order to delete a policy [it must be detached](https://boto3.amazonaws.com/v1/documentation/api/1.28.1/reference/services/iam/client/delete_policy.html)

IAM is limited to [20 transactions per second](https://docs.aws.amazon.com/singlesignon/latest/userguide/limits.html#ssothrottlelimits) (TPS). So, one of two problems would occur when trying to delete. Either it would take a long time to detach the policy from all roles and then delete it, or the throttling limit would be hit and trigger a failure.

For this reason a filler statement will save us from either of these issues at the expense of having a policy with a meaningless statement. Seems like a small and valid price to pay to retain the attachments.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: This is now addressed with a warning system that will attach warnings for any issues encountered, but will still notify that the app configuration update succeeded.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+        # Create the new app deny policy
+        try:
+            deny_policy_statements = []
+            enabled_services_dict = app_configuration.configuration.enabled_services.to_dict()
+            # Check each service
+            for service in enabled_services_dict.keys():
+                # If the service has permissions that need to be denied
+                if service in service_disable_permissions:
+                    # If the service is deactivated
+                    if not enabled_services_dict[service]:
```
AlejandroRigau, line 194: wont this always be false since we are doing `for service in enabled_services_dict.keys():`?

backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -435,21 +444,84 @@ def _detach_iam_policies(self, iam_role_name: str) -> List[str]:
...
+        return json.dumps(policy) if policy is not None else None
+
+    # If the provided policy doesn't exist, it will be created
+    def update_dynamic_policy(
+        self,
+        policy: str,
+        policy_name: str,
+        policy_type: str,
+        policy_type_identifier: str,
+        on_create_attach_to_notebook_role: bool = False,
```
AlejandroRigau, line 478: Dont we want this to always be true? what would be a situation where we wont want to add it to the notebook role?

backend/src/ml_space_lambda/utils/resoruce_utils.py
```diff
@@ -0,0 +1,113 @@
+#
```
AlejandroRigau, line 1: You have a typo in the file name. resoruce => resource

frontend/src/entities/configuration/configuration.tsx
```diff
@@ -57,6 +58,7 @@ export function Configuration () {
             }
         >
             <SpaceBetween direction='vertical' size='xl'>
+                <DynamicConfiguration/>
```
AlejandroRigau, line 61: I think we dont want to add this yet

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+                    if not active_service_in_group and "Statements" in group:
+                        deny_policy_statements.extend(group["Statements"])
+
+            # When all services are activated, apply a filler statement so that the policy does not need to be unattached and deleted
+            # Unattaching from all roles could cause throttling and will require re-attaching to all dynamic roles later
+            if len(deny_policy_statements) == 0:
+                deny_policy_statements.extend(filler_deny_statement)
+
+            # Will create, update the current deny services policy with the new permissions
+            iam_manager.update_dynamic_policy(
```
AlejandroRigau, line 219: I am not super familiar with how this works but what happens if the users bring in their own IAM policies?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: Here is the structure:
```
enabled_services_dict = {
  batchTranslate: True,
  realtimeTranslate: False,
  etc...
}

service_disable_permissions = {
    ServiceType.BATCH_TRANSLATE: {
        "Statements": [
            {
                IAMStatementProperty.EFFECT: IAMEffect.DENY,
            ...
       ]
       "ResourceType": ResourceType.BATCH_TRANSLATE_JOB
  },
  ...
}
```

First we check if there are instructions in `service_disable_permissions` for how to deactivate that service, then we check if the service is deactivated in the newly submitted config `if not enabled_services_dict[service]`. For rexample with 'realtimeTranslate' it would be: `if not enabled_services_dict['realtimeTranslate']` == `if not False` == True

Then if the service is deactivated we will apply the deactivation statements.

I will condense this though into a single conditional and check if the service is deactivated first

LightSeekerSC: Updated: `if not enabled_services_dict[service] and service in service_disable_permissions:`

backend/src/ml_space_lambda/utils/iam_manager.py
LightSeekerSC: Currently all dynamic policies we are creating, we probably do want to attach to the notebook role, but this is siding cautious as we have a limit to the number of policies we can attach and I am attempting to be frugal and not accidentally attach policies we don't intend to by default. The idea is that this should be an intentionally opted in feature as accidentally attaching a policy to the notebook role means it will then be attached to all other dynamic roles in the future.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: Great call! This should all be in a conditional for if they are using dynamic roles. If they are using dynamic roles, then this should always work as expected.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
-    request = json.loads(event["body"])
-    configScope = request["configScope"]
-    version_id = request["versionId"] + 1  # increment so this will be the latest version
+service_disable_permissions = {
+    ServiceType.BATCH_TRANSLATE: {
+        "Statements": [
+            {
+                IAMStatementProperty.EFFECT: IAMEffect.DENY,
+                IAMStatementProperty.ACTION: [
+                    "translate:StopTextTranslationJob",
```
douglas1850, line 76: This applies to everything, but why can't we just say

```
translate:*
```

For all of these things? It seems like there are some actions that we may need to be more fine-grained with, but with something like translate or EMR can we just wildcard everything?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+            version_id=version_id,
+            configuration=new_configuration,
+            changed_by=event["requestContext"]["authorizer"]["principalId"],
+            change_reason=request["changeReason"],
+            created_at=time.time(),
+        )
+
+        try:
+            # Create a record for this configuration and mark it the latest config by incrementing the version Id
+            app_configuration_dao.create(config=app_configuration)
```
douglas1850, line 179: We should probably move all of this logic behind the IAM policy updates; as this is written, we could update the app config, and *then* attempt to update the policy, fail, and have an erroneous config.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+            )
+
+        except Exception as e:
+            logging.exception(f"Failed to update the denied services policy {str(e)}")
+            response_status_code = 207
+            warning_issues.append("Failed when updating the deny policy for deacivating service IAM permissions.")
+
+        # All updates were successfully executed
+        if response_status_code == 200:
+            response_message = f"Successfully updated app configuration"
```
douglas1850, line 234: This is where we can make the app config update call.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+        # App config was successfullyl updated but other elements may have failed resulting in mixed results
+        elif response_status_code == 207:
+            response_message = "Successfully updated app configuration, but some issues were encountered."
```
douglas1850, line 237: An update should be atomic - either everything succeeds, or everything fails. 

The response should be something like:

> An error was encountered while activating/deactivating services. Please contact your system administrator for assistance.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
dustins: Agreed, I have the policy updates happen before the config is persisted in my PR

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+            },
+            {
+                IAMStatementProperty.EFFECT: IAMEffect.DENY,
+                IAMStatementProperty.ACTION: "iam:PassRole",
+                IAMStatementProperty.RESOURCE: get_account_arn("iam", f"role/{IAM_RESOURCE_PREFIX}*"),
+                IAMStatementProperty.CONDITION: {"StringEquals": {"iam:PassedToReception": "translate.amazonaws.com"}},
+            },
+        ],
+    }
+]
```
dustins, line 68: This (and a lot of this file) seems like a lot to push into this file. I [followed your recommendation in my PR](https://github.com/awslabs/mlspace/pull/118/files#diff-ad040d66f148a6020ddc56f75bb1b76358c37a3a0ed2feb2e456f45916ea56a7) and split it out into a helper class so this file isn't so cluttered. I suggest you do the same.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
-    app_configuration = AppConfigurationModel(
-        configScope=configScope,
-        version_id=version_id,
-        configuration=new_configuration,
-        changed_by=event["requestContext"]["authorizer"]["principalId"],
-        change_reason=request["changeReason"],
-        created_at=time.time(),
-    )
 
+@event_wrapper
```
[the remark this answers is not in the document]
LightSeekerSC: The normal @api_wrapper only provides a 200 response with the returned value. @event_wrapper is basically the same, but doesn't assign a 200, so it provides more flexibility for what HTML status code we return

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: Sounds good, I can do that.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+            }
+        ],
+        "ResourceType": ResourceType.LABELING_JOB,
+    },
+    ServiceType.EMR_CLUSTER: {
+        "Statements": [
+            {
+                IAMStatementProperty.EFFECT: IAMEffect.DENY,
+                IAMStatementProperty.ACTION: [
+                    "elasticmapreduce:RunJobFlow",
```
douglas1850, line 114: Can we use `*` for EMR actions?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+        try:
+            # Create a record for this configuration and mark it the latest config by incrementing the version Id
+            app_configuration_dao.create(config=app_configuration)
+        except Exception as e:
+            if e.response["Error"]["Code"] == "ConditionalCheckFailedException":
+                e.response["ResponseMetadata"]["HTTPStatusCode"] = 429
+            raise e
 
-    return f"Successfully updated configuration for {configScope}, version {version_id}."
+        log.info(env_variables)
```
douglas1850, line 185: Errant log

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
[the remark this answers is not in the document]
douglas1850: Are you sure about that? I thought I was returning a different status code than 200 somewhere else. We definitely don't return 200 on exceptions

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+        if env_variables[EnvVariable.MANAGE_IAM_ROLES] and configScope == "global":
+            # Create the new app deny policy
+            try:
+                deny_policy_statements = []
+                enabled_services_dict = app_configuration.configuration.enabled_services.to_dict()
+                log.info(f"Keys: {enabled_services_dict.items()}")
```
douglas1850, line 191: Errant log

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,39 +18,234 @@
...
+                # Check each service
+                for service in enabled_services_dict.keys():
+                    # If the service is deactivated and has instructions on how to deny the service in 'service_disable_permissions'
+                    if not enabled_services_dict[service] and service in SERVICE_DEACTIVATE_PROPERTIES:
+                        if "ResourceType" in SERVICE_DEACTIVATE_PROPERTIES[service]:
+                            suspend_all_of_type(SERVICE_DEACTIVATE_PROPERTIES[service]["ResourceType"])
+                        if "Statements" in SERVICE_DEACTIVATE_PROPERTIES[service]:
```
douglas1850, line 198: Is there ever a scenario where "Statements" wouldn't be one of the properties?

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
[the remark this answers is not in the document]
LightSeekerSC: Yes. [api_wrapper](https://github.com/awslabs/mlspace/blob/main/backend/src/ml_space_lambda/utils/common_functions.py#L100-L113) provides for the binary possibility of a 200 response or a 400 response on exception. It doesn't provide the possibility of other HTTP responses. Since I was doing 207 here, we had to use `event_wrapper`. After other feedback, we may can the `207` since we are pursuing atomic actions, so this could go back to an `api_wrapper`

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: After further discussion with @dustins and @douglas1850 we are going to wait to merge @dustins [PR](https://github.com/awslabs/mlspace/pull/118) and then will move forward with creating a mostly atomic action for handling this.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
LightSeekerSC: No, but just protecting against `KeyErrors`. I guess we could remove this.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
douglas1850: I also just merged my change with the SystemRole so I would advise for leveraging that on the app config update lambda. That way you can stop/terminate/delete resources *after* we've done all of our checks.

LightSeekerSC:
Converting this back to draft while resolving the issue with github actions tests. Utilities were causing issues that were not detected by running `acts` locally.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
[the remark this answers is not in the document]
douglas1850: Following up on this one - sounds like we should revert it to an api wrapper

backend/src/ml_space_lambda/app_configuration/policy_helper/active_service_policy_manager.py
```diff
@@ -0,0 +1,187 @@
...
+                        IAMStatementProperty.ACTION: ["ec2:DescribeInstances", "ec2:DescribeRoutesTables"],
+                        IAMStatementProperty.RESOURCE: "*",
+                    },
+                ],
+                "ResourceType": ResourceType.EMR_CLUSTER,
+            },
+        }
+
+        # Instead of deleting the policy (which requires unattaching from all roles) assign a filler non-impactful statement
+        self.FILTER_DENY_STATEMENTS = [
```
douglas1850, line 130: Is FILTER meant to be FILLER?

backend/src/ml_space_lambda/app_configuration/policy_helper/active_service_policy_manager.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        resource_types_to_suspend = []
+        deny_policy_statements = []
+        enabled_services_dict = app_configuration.configuration.enabled_services.to_dict()
+
+        iam_manager = IAMManager() if iam_manager is None else iam_manager
+
+        # Check each service
+        for service in enabled_services_dict.keys():
+            # If the service is deactivated and has instructions on how to deny the service in 'service_disable_permissions'
```
douglas1850, line 154: Does this comment need to be updated? `service_disable_permissions` only seems to appear in this comment

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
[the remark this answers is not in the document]
LightSeekerSC: In the end we are leaving this as an `event_wrapper` for the following reason:
- Everything good - 200
- Something fails the atomicity - 400+
- Changes applied, but error during resource suspension / cleanup - 207

This gives us bandwidth to handle cases where the `SYSTEM_ROLE` is potentially not handled correctly and helps distinguish it from a complete failure

backend/src/ml_space_lambda/data_access_objects/resource_metadata.py
```diff
@@ -172,6 +172,62 @@ def get_all_for_user_by_type(
...
+    ) -> PagedMetadataResults:
+        # Base values
+        expression_values: Dict[str, Any] = {":resourceType": type}
+        key_condition_expressions = ["resourceType = :resourceType"]
+
+        if project is not None and user is not None:
+            raise ValueError(
+                "Can't use both the 'project' and 'user' parameters. Provide the key_condition as the input parameter and the other as a part of the filter expression and values"
+            )
+        elif project is not None:
```
douglas1850, line 196: More instances of the `is not None` stuff in this file

douglas1850 (approved):

douglas1850:
Either as part of this PR or as a fast-follow-up we need to switch the app config `update` lambda to using the new System Role

backend/src/ml_space_lambda/utils/mlspace_config.py
```diff
@@ -43,30 +44,36 @@ def get_environment_variables() -> dict:
...
+            EnvVariable.NOTEBOOK_ROLE_NAME: os.getenv(EnvVariable.NOTEBOOK_ROLE_NAME, ""),
+            EnvVariable.RESOURCE_METADATA_TABLE: os.getenv(EnvVariable.RESOURCE_METADATA_TABLE, "mlspace-resource-metadata"),
+            EnvVariable.RESOURCE_SCHEDULE_TABLE: os.getenv(EnvVariable.RESOURCE_SCHEDULE_TABLE, "mlspace-resource-schedule"),
+            EnvVariable.PERMISSIONS_BOUNDARY_ARN: os.getenv(EnvVariable.PERMISSIONS_BOUNDARY_ARN, ""),
+            EnvVariable.PROJECTS_TABLE: os.getenv(EnvVariable.PROJECTS_TABLE, "mlspace-projects"),
+            EnvVariable.PROJECT_USERS_TABLE: os.getenv(EnvVariable.PROJECT_USERS_TABLE, "mlspace-project-users"),
+            EnvVariable.S3_KEY: os.getenv(EnvVariable.S3_KEY, "notebook-params.json"),
+            EnvVariable.SYSTEM_TAG: os.getenv(EnvVariable.SYSTEM_TAG, "MLSpace"),
+            EnvVariable.TRANSLATE_DATE_ROLE_ARN: os.getenv(EnvVariable.TRANSLATE_DATE_ROLE_ARN, ""),
+            EnvVariable.USERS_TABLE: os.getenv(EnvVariable.USERS_TABLE, "mlspace-users"),
```
dustins, line 76: I like the change to enums for the environment variables but this has made this section much harder to read. I think we can change it to something like this to clean it up a bit.

```python
[
	(EnvVariable.APP_CONFIGURATION_TABLE, "mlspace-app-configuration")
	(EnvVariable.AWS_DEFAULT_REGION, "us-iso-east-1")
	(EnvVariable.BUCKET, "mlspace-data-bucket")
	(EnvVariable.DATA_BUCKET, "mlspace-data-bucket")
	(EnvVariable.DATASETS_TABLE, "mlspace-datasets")
	(EnvVariable.DYNAMO_TABLE, "mlspace-project")
	(EnvVariable.ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN, "")
	(EnvVariable.EMR_CONFIG_BUCKET, "mlspace-emr-config-bucket")
	(EnvVariable.EMR_EC2_ROLE_NAME, "EMR_EC2_DefaultRole")
	(EnvVariable.EMR_EC2_SSH_KEY, "")
	(EnvVariable.EMR_SERVICE_ROLE_NAME, "EMR_DefaultRole")
	(EnvVariable.EMR_SECURITY_CONFIGURATION, "MLSpace-EMR-SecurityConfig")
	(EnvVariable.JOB_INSTANCE_CONSTRAINT_POLICY_ARN, "")
	(EnvVariable.LOG_BUCKET, "mlspace-log-bucket")
	(EnvVariable.MANAGE_IAM_ROLES, "")
	(EnvVariable.NEW_USER_SUSPENSION_DEFAULT, "True")
	(EnvVariable.NOTEBOOK_ROLE_NAME, "")
	(EnvVariable.RESOURCE_METADATA_TABLE, "mlspace-resource-metadata")
	(EnvVariable.RESOURCE_SCHEDULE_TABLE, "mlspace-resource-schedule")
	(EnvVariable.PERMISSIONS_BOUNDARY_ARN, "")
	(EnvVariable.PROJECTS_TABLE, "mlspace-projects")
	(EnvVariable.PROJECT_USERS_TABLE, "mlspace-project-users")
	(EnvVariable.S3_KEY, "notebook-params.json")
	(EnvVariable.SYSTEM_TAG, "MLSpace")
	(EnvVariable.TRANSLATE_DATE_ROLE_ARN, "")
	(EnvVariable.USERS_TABLE, "mlspace-users")
]

env_variables = dict((env_var, os.getenv(env_var,, default)) for env_var,, default in environment_defaults)

File: backend/src/ml_space_lambda/utils/iam_manager.py [hunk 27]
```diff
@@ -192,6 +192,7 @@ def __init__(self, iam_client=None, sts_client=None):
         env_variables = get_environment_variables()
         self.data_bucket = env_variables["DATA_BUCKET"]
         self.system_tag = env_variables["SYSTEM_TAG"]
+        self.app_role_name = os.getenv("APP_ROLE_NAME", "")
```
dustins, line 195: Should we just add `APP_ROLE_NAME` to the environment variables so it can be read from `env_variables` too? 

Also if we have the enum for environment variables shouldn't we start using them in all the files we touch?

LightSeekerSC: Yes. I scaled this back though because I think `get_environment_variables` was a culprit for mocking issues and I was trying to get around it so I could proceed. I am all for updating

I will try it, if it causes test failures I will let you know and we can revisit it in a separate refactor

AlejandroRigau (commented):
LGTM after you address unresolved conversations and conflicts

LightSeekerSC:
Had to rebase on main again. Made Dustin's requested changes in the newest commit

dustins (approved):

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -110,11 +110,17 @@ export function DynamicConfiguration () {
                 }
             } else {
                 dispatch(getConfiguration({configScope: 'global'}));
-                notificationService.generateNotification(
-                    'Successfully updated configuration.',
-                    'success'
-                );
-                // Increment the versionId so subsequent changes don't fail from "stale" config
```
douglas1850, line 117: My comment 😭

douglas1850 (approved):

AlejandroRigau (approved):

PR #111, opened 2024-05-31
 [AIML-ADC-8475] Configuration history with rollback functionality

AlejandroRigau:
*Issue #, if available:*: AIML-ADC-8475

*Description of changes:*
- Add table with list of previous configurations
- Include button in table to rollback to previous configuration

Screenshot 2024-06-03 at 11 13 13 AM
Screenshot 2024-06-03 at 11 13 29 AM

### DIFF
--- a/frontend/src/entities/configuration/configuration-history-table.tsx
+++ b/frontend/src/entities/configuration/configuration-history-table.tsx
@@ -17,14 +17,14 @@
 import React, { useMemo } from 'react';
 import { useAppDispatch, useAppSelector } from '../../config/store';
 import { IAppConfiguration, defaultConfiguration } from '../../shared/model/app.configuration.model';
-import { appConfig, appConfigList, getConfiguration, listConfigurations, loadingAppConfig, updateConfiguration } from './configuration-reducer';
+import { appConfig, appConfigList, getConfiguration, listConfigurations, loadingAppConfigList, updateConfiguration } from './configuration-reducer';
 import { Box, Button, FormField, Header, Input, Modal, SpaceBetween, Table } from '@cloudscape-design/components';
 import NotificationService from '../../shared/layout/notification/notification.service';
 
 export function ConfigurationHistoryTable () {
     const configList: IAppConfiguration[] = useAppSelector(appConfigList);
     const applicationConfig: IAppConfiguration = useAppSelector(appConfig);
-    const loadingConfig: boolean = useAppSelector(loadingAppConfig);
+    const loadingConfigList: boolean = useAppSelector(loadingAppConfigList);
     const dispatch = useAppDispatch();
     const notificationService = NotificationService(dispatch);
     const [modal, setModal] = React.useState<{ visible: boolean; prevConfig: IAppConfiguration, newConfig: IAppConfiguration }>({
@@ -62,7 +62,7 @@ export function ConfigurationHistoryTable () {
                 header={<Header> Configuration history </Header>}
                 columnDefinitions={columnDefinition}
                 items={configList}
-                loading={loadingConfig}
+                loading={loadingConfigList}
                 loadingText='Loading configurations'
             />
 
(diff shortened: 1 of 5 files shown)


frontend/src/entities/configuration/configuration-reducer.ts
```diff
@@ -28,12 +29,13 @@ import { defaultConfiguration, IAppConfiguration } from '../../shared/model/app.
...
     appConfigList: [],
     appConfig: defaultConfiguration,
+    loadingAppConfig: false,
     failedToLoadConfig: false,
 };
 
 // Actions
 export const getConfiguration = (configScope: string) => {
-    return listConfigurations({ configScope, numVersions: 1});
+    return listConfigurations({ configScope, numVersions: 10});
```
LightSeekerSC, line 38: Don't do this. The intent of this function is to get a single config for normal users. Only admins are supposed to be able to get more than 1 config with the list configurations function.

AlejandroRigau: is there a downside to having all users have a list of 10 configurations in the store rather than just admins? I dont this this information is particularly sensitive since users had access to it anyways.

frontend/src/entities/configuration/configuration-history-table.tsx
```diff
@@ -0,0 +1,107 @@
...
+                                        'success'
+                                    );
+                                }
+                            }
+                            }>Rollback</Button>
+                        </SpaceBetween>
+                    </Box>
+                }
+            >
+                Are you sure you want to rollback to version {modal.prevConfig.versionId}?
```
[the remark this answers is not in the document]
AlejandroRigau: I can add a text box in the modal to add an optional message

frontend/src/entities/configuration/configuration-reducer.ts
LightSeekerSC: @douglas1850 would have the best answer. My understanding is that general users shouldn't need anything other than the current. I don't think there is a strict reason why this would be a security issue, but limiting to one was preferred in the original design.

douglas1850: The configs aren't sensitive information, but there's no reason to grab 10 when a typical user would only ever need 1. We're 10x'ing the number of reads from DDB for absolutely no reason. 

If you're not sure how to fetch configs, you can invoke the `listConfigurations` function directly like this:

```
dispatch(listConfigurations({configScope: 'global', numVersions: 10}));
```

AlejandroRigau: The problem with calling listConfigurations directly is that `appConfigList: data,` gets overwritten to only display 1 configuration. This is because we call getConfiguration in the routes component and then when I call listConfigurations in the history table component, it also gets set to num versions 1. I can probably pass an extra parameter to listConfigurations when calling it from getConfiguration so that we dont update appConfigList in that call

douglas1850: We originally had separate reducers for `get` and `list` but switched to doing it this way to minimize repetitive code. 

It seems like this might be resolved by simply separating those out again. If `get` and `list` had its own reducer then we could make sure that the `get` calls weren't setting the appConfigList field, and that `list` calls don't set the appConfig field.

AlejandroRigau:
Screenshot 2024-06-03 at 3 36 15 PM

New modal accepting change reason

frontend/src/entities/configuration/configuration-reducer.ts
```diff
@@ -78,15 +88,44 @@ export const AppConfigSlice = createSlice({
...
                     failedToLoadConfig: true,
+                    loadingAppConfig: false,
+                };
+            })
+            .addMatcher(isFulfilled(getConfiguration), (state, action: any) => {
+                const { data } = action.payload;
+                return {
+                    ...state,
+                    appConfig: data[0],
+                    loadingAppConfig: false,
```
douglas1850, line 115: Should `loadingAppConfig` be used for `getConfiguration` too? It looks like we use it for both get and list, so if one starts loading this will be true, and if one finishes loading before the other then this will be false for both. 

It looks like you're specifically using this for list, so perhaps just change it to `loadingAppConfigList` and only use it for the list call.

douglas1850 (approved):

LightSeekerSC (approved):

PR #112, opened 2024-06-05
[AIML-ADC-8474] Display changes to config file before submitting

AlejandroRigau:
*Issue #, if available:* AIML-ADC-8474

*Description of changes:*
- Adds function to compare previous config to new config.
- Adds modal displaying changes before submitting
- Automatically add a description of changes based on changed object keys

Screenshot 2024-06-05 at 11 36 09 AM

### DIFF
--- a/frontend/src/entities/configuration/dynamic-configuration.tsx
+++ b/frontend/src/entities/configuration/dynamic-configuration.tsx
@@ -28,6 +28,7 @@ import {
     ContentLayout,
     Modal,
     Box,
+    TextContent,
 } from '@cloudscape-design/components';
 import React, { useEffect, useMemo, useState } from 'react';
 import { useAppDispatch, useAppSelector } from '../../config/store';
@@ -41,6 +42,7 @@ import { formatDisplayNumber } from '../../shared/util/form-utils';
 import { ClusterTypeConfiguration } from './cluster-types';
 import { InstanceTypeMultiSelector } from '../../shared/metadata/instance-type-dropdown';
 import { ConfigurationHistoryTable } from './configuration-history-table';
+import _ from 'lodash';
 
 export function DynamicConfiguration () {
     const applicationConfig: IAppConfiguration = useAppSelector(appConfig);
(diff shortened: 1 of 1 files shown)


frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -218,413 +223,520 @@ export function DynamicConfiguration () {
...
-                        variant='default' 
-                        headingTagOverride='h3' 
-                        headerDescription='A list of applications for Amazon EMR to install and configure when launching the cluster.'
+                    </Box>
+                }
+            >
+                <SpaceBetween size={'s'}>
+                    <Container>
+                        <TextContent>
+                            {_.isEmpty(changesDiff) ? <p>No changes detected</p> : jsonToOutline(changesDiff)}
```
douglas1850, line 310: Can we use this same logic to disable the "Save Changes" button so that we're only enabling it when the user has made changes?

AlejandroRigau: I figured this way the user would at least know why the button is deactivated.

douglas1850: I'm talking about the Save changes button on line 671 where we determine if that button is disabled or not - can we apply this same logic there? So the button is disabled if:

* the form is submitting
* the form is invalid 
* no changes have yet been made

AlejandroRigau: if I disable that button, the user wont know why its disabled. I made it so that you could still click on it but then it shows the modal like this:

Screenshot 2024-06-05 at 12 12 32 PM

So the user will still not be allowed to submit without changes but will have a visual indication as to why it cant. Ultimately Ill change it to whatever you think makes more sense.

douglas1850: You know what I like that better than what I was thinking, let's keep that

douglas1850 (dismissed):

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -218,413 +223,520 @@ export function DynamicConfiguration () {
...
+     * Computes the difference between two JSON objects, recursively.
+     *
+     * This function takes two JSON objects as input and returns a new object that
+     * contains the differences between the two. Works with nested objects.
+     *
+     * @param {object} [obj1={}] - The first JSON object to compare.
+     * @param {object} [obj2={}] - The second JSON object to compare.
+     * @returns {object} - A new object containing the differences between the two input objects.
+     */
+    function getJsonDifference (obj1 = {}, obj2 = {}) {
```
LightSeekerSC, line 239: If we are making this as a utility it should be in a utility file.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -218,413 +223,520 @@ export function DynamicConfiguration () {
...
+        const output = {},
+            merged = { ...obj1, ...obj2 }; // has properties of both
+
+        for (const key in merged) {
+            const value1 = obj1[key], value2 = obj2[key];
```
LightSeekerSC, line 244: If this is a utility it should account for the fact that the key may not exist. This expects both objects to have the same top-level keys

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -218,413 +223,520 @@ export function DynamicConfiguration () {
...
+
+            if (isObject(value1) || isObject(value2)) {
```
LightSeekerSC, line 246: This works currently because both objects have the same structure, but this should probably be an `&&` instead of a `||`. This function expects both inputs to be objects and I think would have an error if you attempted `merged = { ...obj1, ...obj2};` if one of them is not a dict/object

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -218,413 +223,520 @@ export function DynamicConfiguration () {
...
+                const value = getJsonDifference(value1, value2); // recursively call
+                if (Object.keys(value).length !== 0) {
+                    output[key] = value;
+                }
+
+            } else {
+                if (JSON.stringify(value1) !== JSON.stringify(value2)) {
```
LightSeekerSC, line 253: `~.isEqual`may be preferred here: https://lodash.com/docs/4.17.15#isEqual

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -218,413 +223,520 @@ export function DynamicConfiguration () {
...
+                    output[key] = value2;
```
LightSeekerSC, line 254: Alternatively, here we could add a new object of the following structure:
```
key: {
  a: value1,
  b: value2
}
```

This makes it possible to visualize the difference of what is being changed. a->b

LightSeekerSC (commented):
The current structure with the `<ul>` has pretty unnecessary line-height. Is there a way to reduce this further?

It would be nice to see what changes are occurring. Either `a -> b` or updating with new value: `b`

I wouldn't block on showing the diff, but do think we need to reduce the line-height

AlejandroRigau:
> The current structure with the `<ul>` has pretty unnecessary line-height. Is there a way to reduce this further?
> 
> It would be nice to see what changes are occurring. Either `a -> b` or updating with new value: `b`
> 
> I wouldn't block on showing the diff, but do think we need to reduce the line-height

@LightSeekerSC it is using the standard cloudscape determined line heights that are determined as part of the theme, I think. Any suggestion as to what else I could use other than ul?

Also, I did try with the values and had a discussion on slack about it. We determined that adding the values makes it too messy. 

See: 
image (1)

frontend/src/entities/configuration/dynamic-configuration.tsx
AlejandroRigau: I wasn't really planning on making this a utility. its a very specific usecase for this file. I only made it into a function so that its cleaner. I wont move it to a utility file unless you think this file is getting out of hand.

frontend/src/entities/configuration/dynamic-configuration.tsx
AlejandroRigau: it already does handle that. if they key does not exist, it returns undefined which everything works normally if its undefined. `isObject(value1)` would just return `false` and then `JSON.stringify(value1)` would make undefined into a string 'undefined' so it all works normally. When we recursively call `getJsonDifference(value1, value2)` with an undefined, the return is always an object so the `merged = { ...obj1, ...obj2};` part will work fine.

```
console.log(getJsonDifference({test: 'wa', test2: 'na'}, {test: 'wa'}));

>> {test2: undefined}
```
Which is the expected output.

frontend/src/entities/configuration/dynamic-configuration.tsx
AlejandroRigau: same as above

frontend/src/entities/configuration/dynamic-configuration.tsx
AlejandroRigau: same as above. We had decided it would be too cluttered to add the differences.

frontend/src/entities/configuration/dynamic-configuration.tsx
AlejandroRigau: why? `JSON.stringify(value1)` seems to handle all situations we would face. Is there any benefit to using lodash?

LightSeekerSC: Generally lodash's libraries are optimized for a variety of edge cases as well as performance.

For primitives it doesn't matter much, but lodash is equal is better in any scenarios where this may be an array or more complex object: https://www.samanthaming.com/tidbits/33-how-to-compare-2-objects/

LightSeekerSC:
@AlejandroRigau Would [in-line styles](https://www.codecademy.com/article/html-inline-styles) work? It appears you are using a normal `<ul>` so in-line styles will override the class.

I think the CSS for [line-height](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height) should work

douglas1850:
Personally I think the line height is okay, I'd just make sure the modal will elegantly handle situations where a customer makes changes to *all* fields and the modal scrolls/resizes to adjust for this.

I do think mapping the property names to human-readable names could be a good idea.

AlejandroRigau:
I want to avoid inline style for things like text spacing. text spacing should be consistent across the app and would be helpful when theming. If someone where to change that value in a theme, this section would be the only one to not get applied.

AlejandroRigau:
Just spoke to Cindy about it. She likes it how it is right now so that we maintain consistent spacing with the rest of the theme

douglas1850 (approved):

LightSeekerSC (approved):

PR #113, opened 2024-06-05
Populate instance types

douglas1850:
*Description of changes:* Create a temporary policy that the custom CDK resource uses to populate the active instance types.

The lambda function (and this policy) are deleted after deployment is complete. 

### DIFF
--- a/backend/src/ml_space_lambda/data_access_objects/app_configuration.py
+++ b/backend/src/ml_space_lambda/data_access_objects/app_configuration.py
@@ -393,3 +393,14 @@ class AppConfigurationDAO(DynamoDBObjectStore):
             scan_index_forward=False,
         ).records
         return json_response
+
+    # Currently only used to update the initial app config (versionId=0)
+    def update(self, config: dict) -> None:
+        json_key = {"configScope": "global", "versionId": 0}
+        update_exp = "SET configuration = :config"
+        exp_values = json.loads(dynamodb_json.dumps({":config": config}))
+        self._update(
+            json_key=json_key,
+            update_expression=update_exp,
+            expression_values=exp_values,
+        )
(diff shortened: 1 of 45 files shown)


LightSeekerSC:
This faces the issue with customers providing manual IAM resources needing to pre-create this as part of install. Can this just use the app policy instead? It should be a part of the app role

AlejandroRigau:
^^ same as LightSeekerSC

douglas1850:
This is replacing the app role because using the app role broke it. 

This is creating an IAM policy for a lambda function that is only executed onCreate, and then the lambda and its policy are deleted.

douglas1850:
We're doing something similar in this file, fwiw:

```
// Populate the App Config table with default config
        new AwsCustomResource(this, 'mlspace-init-ddb-app-config', {
            onCreate: {
                service: 'DynamoDB',
                action: 'putItem',
                parameters: {
                    TableName: props.mlspaceConfig.APP_CONFIGURATION_TABLE_NAME,
                    Item: generateAppConfig(props.mlspaceConfig),
                },
                physicalResourceId: PhysicalResourceId.of('initAppConfigData'),
            },
            policy: AwsCustomResourcePolicy.fromSdkCalls({ resources: [appConfigTable.tableArn] }),
        });
```

AlejandroRigau (approved):

AlejandroRigau:
I approved but where does it delete it after create?

douglas1850:
Hmm actually I was mistaken, it deletes resources for something like the PutItem example I posted, but this lambda will persist since I defined it myself. Will look into trying to get the app config role to work.

estohlmann (approved):

douglas1850:
I'm merging this in for now so deployments with the latest code won't fail. 

I'm meeting with the CDK team tomorrow during their office hours to get to the bottom of this. If we can figure something out using the mlspaceAppRole I'll implement that, if we can't figure something out then I'll rip out this functionality and we'll decide on a different path forward. 

Our options if the CDK team can't help me find a solution are:
1. leave the enabled instances list blank by default and tell admins to select instances they want enabled during setup
2. curate a list of instance types available in all regions so there's a few instances in the list as a starting point

PR #118, opened 2024-06-07
Backend updates for notebook instance constraints

dustins:
*Issue #, if available:*

*Description of changes:*
Updates to the backend to work with new application configuration instance type constraints on notebooks.

**CDK**
  - updated notebook role policy to no longer `ALLOW` Training/HPO/Transform/EndpointConfig create by default
  - added two new policies to the notebook role that only allow Training/HPO/Transform/EndpointConfig create when using configured instance types

**Backend**
- updated app config lambda to update notebook instance type constraint policies when updating


lib/constants.ts
```diff
@@ -107,6 +107,12 @@ export const ENABLE_GROUNDTRUTH = true;
 // The default name for the application
 export const APPLICATION_NAME = 'MLSpace';
 
+// Policy names attached to NOTEBOOK_ROLE_ARN that restricts instance types that a notebook
+// can use for each service
+export const ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN = '';
```
AlejandroRigau, line 112: I thought we were hoping to deprecate constants.ts. Should we avoid adding new things here?

backend/src/ml_space_lambda/app_configuration/policy_helper/notebook.py
```diff
@@ -0,0 +1,90 @@
...
+        "Condition": {"ForAnyValue:StringEquals": {"sagemaker:InstanceTypes": allowed_instances}},
+    }
+
+
+def create_sagemaker_resource_arn(resource: str, context) -> str:
+    components = context.invoked_function_arn.split(":")
+    return f"arn:{components[1]}:sagemaker:{components[3]}:{components[4]}:{resource}/*"
+
+
+def update_instance_constraint_policies(new_configuration, context) -> None:
```
douglas1850, line 43: in this function I see it handles:

* Training/HPO
* Transform
* Endpoints

Do we need to handle Notebooks here too? That would be the last configurable service with instance types

dustins: I'll check but I thought we don't allow notebooks to launch other notebooks.

backend/test/app_configuration/test_update_configuration.py
```diff
@@ -170,3 +186,86 @@ def test_update_config_unexpected_exception(mock_app_config_dao):
 
     mock_app_config_dao.create.side_effect = ClientError(error_msg, "PutItem")
     assert lambda_handler(mock_event, mock_context) == expected_response
+
+
+@mock.patch.dict("os.environ", TEST_ENV_CONFIG, clear=True)
+@mock.patch("ml_space_lambda.app_configuration.policy_helper.notebook.iam")
+def test_update_instance_constraint_policies(iam):
```
LightSeekerSC, line 193: For testing organization, I think we should either handle this within the context of a valid app config update or move this test to a separate file I think.

I think if we are moving all of this to a separate file to de-clutter the app-config then it probably does make sense for us to move this to a separate test file too. If we are testing it as part of a call to update then I feel it makes more sense to keep here

Won't block on this. Choose what you think makes sense.

lib/constants.ts
douglas1850: We're migrating *some* things from constants.ts, but this entire file won't be done away with. 

The idea is that in the future this file will only be deployment-time configurations, and then things like ENABLE_TRANSLATE and ENABLE_GROUNDTRUTH will be handled by app config.

So anything dealing with roles/policies will need to remain in here.

backend/src/ml_space_lambda/app_configuration/policy_helper/notebook.py
douglas1850: Ah okay I missed that detail, I thought this was app-wide, not specific to notebooks. This looks good then

lib/stacks/api/appConfiguration.ts
```diff
@@ -54,6 +54,10 @@ export class AppConfigurationApiStack extends Stack {
                 description: 'Update the MLSpace application configuration',
                 path: 'app-config',
                 method: 'POST',
+                environment: {
+                    ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN: props.endpointConfigInstanceConstraintPolicy.managedPolicyArn,
```
LightSeekerSC, line 58: Why are we doing this here? Shouldn't this be populated in the env_variables from the `mlspace_config`?

lib/stacks/iam.ts
```diff
@@ -195,23 +201,23 @@ export class IAMStack extends Stack {
                 }),
                 // Endpoint Configuration Permissions
                 new PolicyStatement({
-                    effect: Effect.ALLOW,
+                    effect: Effect.DENY,
```
LightSeekerSC, line 204: Need to make updates to the [docs](https://github.com/awslabs/mlspace/blob/main/frontend/docs/admin-guide/install.md?plain=1#L213-L225) for these changes.

This is actually going to complicate user installation, because these shouldn't be changed if the customer is not using dynamic roles (most are though)

If not using dynamic roles (few users): Use `Effect.ALLOW` with previous permissions
If using dynamic roles (`MANAGE_IAM_ROLES`): Use these `Effect.DENY` statements

In the docs we will have to provide clear instructions to installers on how to apply the appropriate permissions.

lib/constants.ts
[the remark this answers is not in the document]
dustins: Yeah this is partially why I wanted to sync up about the role/policy permissions. I'll add a child task under the docs for this.

lib/stacks/iam.ts
```diff
@@ -396,12 +405,65 @@ export class IAMStack extends Stack {
...
+            ];
+        };
+
         const notebookPolicy = new ManagedPolicy(this, 'mlspace-notebook-policy', {
             statements: notebookPolicyStatements(this.partition, Aws.REGION),
             description: 'Enables general MLSpace actions in notebooks and across the entire application.'
         });
 
-        // If roles are manually created use the existing role
+        if (props.mlspaceConfig.ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN) {
```
LightSeekerSC, line 437: This if-else conditional and the other should be within a conditional checking for `MANAGE_IAM_ROLES`. These policies can't / shouldn't be created if dynamic roles are not being used.

lib/stacks/iam.ts
```diff
@@ -422,7 +484,11 @@ export class IAMStack extends Stack {
             this.mlSpaceNotebookRole = new Role(this, 'mlspace-notebook-role', {
                 roleName: mlSpaceNotebookRoleName,
                 assumedBy: notebookPolicyAllowPrinciples,
-                managedPolicies: [notebookPolicy],
+                managedPolicies: [
+                    notebookPolicy,
+                    this.mlspaceEndpointConfigInstanceConstraintPolicy,
```
LightSeekerSC, line 489: Also should only be added if `MANAGE_IAM_POLICIES` is set

lib/utils/configTypes.ts
```diff
@@ -174,6 +179,8 @@ export function generateConfig () {
         // Notebook settings
         MLSPACE_LIFECYCLE_CONFIG_NAME: MLSPACE_LIFECYCLE_CONFIG_NAME,
         NOTEBOOK_PARAMETERS_FILE_NAME: NOTEBOOK_PARAMETERS_FILE_NAME,
+        ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN: ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN,
```
LightSeekerSC, line 182: These need to be added to `bin/scripts/config-helper` most likely.

backend/src/ml_space_lambda/app_configuration/lambda_functions.py
```diff
@@ -18,20 +18,30 @@
...
 
+from ml_space_lambda.app_configuration.policy_helper.notebook import update_instance_constraint_policies
 from ml_space_lambda.data_access_objects.app_configuration import AppConfigurationDAO, AppConfigurationModel, SettingsModel
 from ml_space_lambda.utils.common_functions import api_wrapper
 
 log = logging.getLogger(__name__)
 app_configuration_dao = AppConfigurationDAO()
 
 
+@api_wrapper
```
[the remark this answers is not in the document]
dustins: No, I think I accidentally just pushed this around while I was adding stuff to the file and now that moved everything out of this file. I can see about moving this back to the original order.

backend/test/app_configuration/test_update_configuration.py
dustins: I think splitting this out to its own test file probably makes more sense.

lib/stacks/iam.ts
dustins: Good callout although I'm not 100% sure it needs to be checking for `MANAGE_IAM_ROLES`. My understanding was `MANAGE_IAM_ROLES` was for dynamic user roles not just anywhere we're updating roles dynamically.

I'm going to move these two conditions inside the else condition below for when the notebook role isn't explicitly defined. This will then follow the same pattern as was already being done for the notebook role.

lib/stacks/iam.ts
dustins: I'll wait to see if you agree with my comment above. I think if we're creating the notebook role these are being applied regardless of if they're predefined or we're creating them during cdk deploy.

lib/stacks/iam.ts
dustins: Based on what Doug mentioned earlier I added a task for documenting this. This will be hard to explain simply but great callout!

lib/stacks/iam.ts
dustins: After speaking offline I see why this is necessary and will make these changes.

lib/stacks/api/appConfiguration.ts
douglas1850: In `lib/utils/apiFunction` there's this block:

```
environment: {
            DATASETS_TABLE: mlspaceConfig.DATASETS_TABLE_NAME,
            PROJECTS_TABLE: mlspaceConfig.PROJECTS_TABLE_NAME,
            PROJECT_USERS_TABLE: mlspaceConfig.PROJECT_USERS_TABLE_NAME,
            USERS_TABLE: mlspaceConfig.USERS_TABLE_NAME,
            RESOURCE_SCHEDULE_TABLE: mlspaceConfig.RESOURCE_SCHEDULE_TABLE_NAME,
            RESOURCE_METADATA_TABLE: mlspaceConfig.RESOURCE_METADATA_TABLE_NAME,
            APP_CONFIGURATION_TABLE: mlspaceConfig.APP_CONFIGURATION_TABLE_NAME,
            SYSTEM_TAG: mlspaceConfig.SYSTEM_TAG,
            MANAGE_IAM_ROLES: mlspaceConfig.MANAGE_IAM_ROLES ? 'True' : '',
            NOTEBOOK_ROLE_NAME: notebookRoleName,
            PERMISSIONS_BOUNDARY_ARN: permissionsBoundaryArn || '',
            ...funcDef.environment,
            ...mlspaceConfig.ADDITIONAL_LAMBDA_ENVIRONMENT_VARS,
        },
```

Where we give most lambdas some baseline env vars... but since these only apply to this lambda, this seems like the perfect place for them to be defined.

lib/stacks/iam.ts
LightSeekerSC: These do need to be conditionally added

dustins: Yep based on what we talked about I made those changes along with responding to your other comments

lib/stacks/iam.ts
LightSeekerSC: In theory, the original `Effect.ALLOW` style permissions should be applied if `MANAGE_IAM_ROLES` is false, but this is only for the case that someone has the permissions to do a CDK deploy that will create these IAM roles and policies and then is intentionally turning off `MANAGE_IAM_ROLES`. I think this should never be a scenario that happens, but is a possible edge-case.

I won't block on this though, because this is a huge anti-pattern no one should probably do

backend/test/app_configuration/policy_helper/test_notebook.py
```diff
@@ -0,0 +1,122 @@
...
+        {
+            ServiceType.NOTEBOOK.value: ["ml.t3.medium"],
+            ServiceType.ENDPOINT.value: ["ml.t3.large"],
+            ServiceType.TRAINING_JOB.value: ["ml.t3.xlarge"],
+            ServiceType.TRANSFORM_JOB.value: ["ml.t3.kindabig"],
+        }
+    )
+
+    update_instance_constraint_policies(new_configuration, mock_context)
+    iam.create_policy_version.assert_has_calls(
```
LightSeekerSC, line 58: If this is just intended to test the update function in isolation, then the mock should probably be of the `create_instance_constraint_policy_version` function. If this is intended to test overall success it seems that it would make sense ensuring the proper statements were used for the call instead of `mock.ANY`.

Won't block, because this is mostly covered already.

backend/test/app_configuration/policy_helper/test_notebook.py
```diff
@@ -0,0 +1,122 @@
...
+        PolicyDocument=json.dumps(
+            {
+                "Version": "2012-10-17",
+                "Statement": statements,
+            }
+        ),
+        SetAsDefault=True,
+    )
+
+    iam.create_policy_version.assert_called_once()
```
LightSeekerSC, line 121: Duplicate line with no further processing. Should be removed.

bin/scripts/config-helper.mjs
```diff
@@ -73,6 +73,15 @@ async function advancedConfigPrompts () {
         await askRoleQuestions();
     }
 
+    const policiesResponse = await prompt({
```
LightSeekerSC, line 76: I would just update `existingRoles` to say "use existing IAM Roles/Policies?". This won't be necessary for users not using Dynamic Roles, but I would just say it is optional and still group them together as it all is in the vein of IAM resources

lib/stacks/iam.ts
```diff
@@ -396,19 +405,79 @@ export class IAMStack extends Stack {
...
+                }),
+            ];
+        };
+
         const notebookPolicy = new ManagedPolicy(this, 'mlspace-notebook-policy', {
             statements: notebookPolicyStatements(this.partition, Aws.REGION),
             description: 'Enables general MLSpace actions in notebooks and across the entire application.'
         });
 
-        // If roles are manually created use the existing role
```
LightSeekerSC, line 404: Why remove this comment?

lib/stacks/iam.ts
```diff
@@ -396,19 +405,79 @@ export class IAMStack extends Stack {
...
         if (props.mlspaceConfig.NOTEBOOK_ROLE_ARN) {
             this.mlSpaceNotebookRole = Role.fromRoleArn(
                 this,
                 'mlspace-notebook-role',
                 props.mlspaceConfig.NOTEBOOK_ROLE_ARN
             );
         } else {
+            const managedPolicies: IManagedPolicy[] = [notebookPolicy];
+
+            if (props.mlspaceConfig.MANAGE_IAM_ROLES) {
```
LightSeekerSC, line 446: This should not be here. Otherwise this will not trigger during a deployment where the user is manually providing the `NOTEBOOK_ROLE_ARN`. This should be independent of the `NOTEBOOK_ROLE_ARN` check.

lib/stacks/iam.ts
dustins: Nope, no idea how that got removed. I'll add it back.

backend/test/app_configuration/policy_helper/test_notebook.py
```diff
@@ -0,0 +1,115 @@
...
+        {
+            ServiceType.NOTEBOOK.value: ["ml.t3.medium"],
+            ServiceType.ENDPOINT.value: ["ml.t3.large"],
+            ServiceType.TRAINING_JOB.value: ["ml.t3.xlarge"],
+            ServiceType.TRANSFORM_JOB.value: ["ml.t3.kindabig"],
+        }
+    )
+
+    update_instance_constraint_policies(new_configuration, mock_context)
+    create_instance_constraint_policy_version.assert_has_calls(
```
LightSeekerSC, line 58: I think this should be updated in some way to evaluate the combining of the policies, but I am not going to block on it now, since I need this to proceed.

LightSeekerSC (approved):

AlejandroRigau (dismissed):

douglas1850 (approved):

estohlmann (dismissed):

estohlmann (approved):

douglas1850 (approved):

PR #120, opened 2024-06-08
[AIML-ADC-8348] Provide more user friendly error messages for quota issues or name conflicts

AlejandroRigau:
*Issue #, if available:* AIML-ADC-8348

*Description of changes:*
- Provide more user friendly error messages for quota issues or name conflicts (all in a central location)
- Updates to reducer to capture errors
- Fix error response notification
- Fix model notification that incorrectly referenced "endpoint"

Tested name conflicts with training, labeling, endpoint, endpoint config, models, notebooks.
Tested quotas by creating two ml.m5.12xlarge notebooks. Account limit is 1. (yes, I turned them off afterwards)

Screenshot 2024-06-07 at 9 00 16 PM


AlejandroRigau:
I have a bug where start/create notebooks errors are handled differently. Need to find where the differences are and make them the same

backend/src/ml_space_lambda/utils/common_functions.py
```diff
@@ -200,16 +200,23 @@ def generate_html_response(status_code, response_body):
...
         metadata = e.response.get("ResponseMetadata")
         if metadata:
             status_code = metadata.get("HTTPStatusCode", 400)
+        if e.response["Error"]["Code"] in ["ResourceInUse", "ValidationException"] and any(
+            error in error_msg for error in ["Cannot create a duplicate", "already existing", "already exists"]
+        ):
+            e = f"The resource name you provided already exists. Please choose a different name. Full message: {error_msg}"
+        elif e.response["Error"]["Code"] in ["ResourceLimitExceeded"]:
+            e = f"You have reached the maximum allowed usage for this resource. Please contact your MLSpace administrator to increase the allowed usage limits. Full message: {error_msg}"
+
```
dustins, line 214: This is great! It might make sense to break out these errors though so it is more cleanly extensible in the future. Then we could just loop over our known errors a little more cleanly. Then our error checking logic never gets more complicated, we just add to our `ERROR_MAP` structure.

```python
ERROR_MAP = [{
  Type: "duplicate"
  Codes: ["ResourceInUse", "ValidationException"],
  MatchStrings: ["Cannot create a duplicate", "already existing", "already exists"],
  FriendlyMessage: "The resource name you provided already exists. Please choose a different name."
 }, {
  Type: "limit"
  Codes: ["ResourceLimitExceeded"],
  # hack to always match, could add logic to check if MatchStrings is empty or not
  MatchStrings: [""],
  FriendlyMessage: "You have reached the maximum allowed usage for this resource. Please contact your MLSpace administrator to increase the allowed usage limits."
}]
...

for errorType in ERROR_MAP:
  if e.response["Error"]["Code"] in errorType["Codes"] and any(error in error_msg for error in errorType["MatchStrings"]):
    e = f"{errorType["FriendlyMessage"]} Full message: {error_msg}"
```

frontend/src/entities/model/create/model-create.tsx
```diff
@@ -162,14 +162,14 @@ export function ModelCreate () {
                         navigate(`/project/${projectName}/model/${state.form.ModelName}`);
                     } else {
                         notificationService.generateNotification(
-                            `Failed to create model because: ${response.data}`,
+                            `Failed to create model with error: ${response.data}`,
```
dustins, line 165: What are your thoughts about removing the `with error` from these types of messages? It seems redundant after saying it failed and adds extra text to what already could be a long message.

frontend/src/entities/notebook/notebook.actions.tsx
```diff
@@ -198,7 +198,7 @@ const NotebookActionHandler = async (
             `${success ? 'Successfully' : 'Failed to'} ${e.detail.id}${
                 e.detail.id === 'stop' ? 'p' : ''
             }${success ? 'ing' : ''} notebook instance ${notebook.resourceId}${
-                success ? '.' : ` because: ${response.data}`
+                success ? '.' : ` because: ${response.payload}`
             } `,
```
dustins, line 202: I see this message isn't your doing but this error message is incredibly unreadable IMHO. I think we should break out some of this inline logic into their own variables (or just use an if-statement for success or not) so this string because easier to visually parse. If you agree lets also stop checking success in such a manual way and instead use `isFulfilled` ([link](https://github.com/awslabs/mlspace/blob/main/frontend/src/entities/batch-translate/create/batch-translate-create.tsx#L203) to how it is used in our code already).

frontend/src/entities/model/create/model-create.tsx
AlejandroRigau: I should probably just revert this. We use the `because` wording everywhere else

backend/src/ml_space_lambda/utils/common_functions.py
AlejandroRigau: I like it! Ill add this

dustins (approved):

estohlmann (approved):

PR #121, opened 2024-06-10
System role

douglas1850:
*Description of changes:* this PR introduces the MLSpace System Role which exists to clean up resources associated with suspend/deleted projects as well as suspended users. In the future this role can be leveraged for any actions that MLSpace as a system should be capable of performing but which shouldn't be permissible for a user to do.

List of changes:
* introduction of the MLSpace system role
* fixed the populateAllowedInstances custom CDK resource - it will now make use of the mlspaceAppRole properly
* fixed a bug for the populateAllowedInstances custom CDK resource where it would overwrite the createdAt, changedBy, and changeReason fields with empty strings
* during testing I found out that certain EMR applications don't deploy correctly, so I'm removing them

**NOTE**: if the large number of changes in `iam.ts` is making your head spin, worry not - the only *actual* change is that I moved all of the app policy statements into their own function (just like we do for the notebook policy), and then at the bottom I created a new section for the system role. No policy statements were added or removed in the making of this PR.

### DIFF
--- a/.devcontainer/post_create_command.sh
+++ b/.devcontainer/post_create_command.sh
@@ -7,6 +7,8 @@ npm install
 
 python3 -m venv .venv
 . .venv/bin/activate
+echo ". .venv/bin/activate" >> ~/.bashrc
+echo ". .venv/bin/activate" >> ~/.zshrc
 cd backend
 pip install .
 pip3 install -r requirements.txt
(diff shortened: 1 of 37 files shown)


AlejandroRigau (approved):
@douglas1850 and I (and the whole team) had a conversation about this PR. System role will have more permissions than users and we agree that it makes sense.

PR #123, opened 2024-06-10
fix default value of modal input field to be the form change reason

AlejandroRigau:
*Issue #, if available:*

*Description of changes:*
Fix for two different bugs:
- Updating the configuration was not updating the configuration table histroy automatically.
- When saving changes in admin config, the expected changeReason is now set

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #124, opened 2024-06-11
Refactor configuration UI

estohlmann:
AIML-ADC-9145

- Refactored the admin ui to be more componentized and take advantage of the entire screen.
- Updated some to the language with PM approved content
- Broke components down to be spread across tabs and various modals so the screen isn't so huge
- Added expand all functionality


frontend/src/entities/configuration/allowed-instance-types-configuration.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const selectedEndpointInstanceOptions = useMemo(() => props.enabledEndpointInstanceTypes.map((instance) => {
+        return {
+            value: instance,
+            label: instance,
+        };
+    }), [props.enabledEndpointInstanceTypes]);
+
+    return (
+        <>
+            <Container
```
dustins, line 70: Is there a point to using `<>` when `<Container />` is the only child?

frontend/src/entities/configuration/activated-services-configuration.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const configurableServices = {
+    batchTranslate: 'Amazon Translate batch',
+    realtimeTranslate: 'Amazon Translate real-time',
+    emrCluster: 'Amazon EMR',
+    labelingJob: 'Amazon Ground Truth'
+};
+
+export type ActivatedServicesConfigurationProps = {
+    setFields: (any) => void;
```
dustins, line 33: You don't have to change this here but since we use things like this `setFields` function all over the place maybe in the future we should have dedicated type definition we can use.

```typescript
type StringMap = {[key: string]: string};
```

frontend/src/entities/configuration/configuration-import-modal.tsx
```diff
@@ -0,0 +1,95 @@
...
+                        >
+                            Import Configuration
+                        </Header>
+                    }
+                >
+                    <SpaceBetween direction='vertical' size='s'>
+                        <FileUpload
+                            onChange={({ detail }) => {
+                                props.setSelectedFile([]); // ensure there's never more than one file
+                                props.setSelectedFile(detail.value);
```
dustins, line 73: Why are these two calls necessary? Won't this second call just always override the first since `detail.value` is `File[]`

frontend/src/entities/configuration/system-banner-configuration.tsx
```diff
@@ -0,0 +1,120 @@
...
+} from '@cloudscape-design/components';
+import React from 'react';
+
+export type SystemBannerConfigurationProps = {
+    setFields: (any) => void;
+    textColor: string;
+    backgroundColor: string;
+    text: string;
+    isEnabled: boolean;
+    touchFields: (any) => void;
```
dustins, line 34: We also have a type definition for [touchFields](https://github.com/awslabs/mlspace/blob/main/frontend/src/shared/validation/index.ts#L134)

frontend/src/entities/configuration/emr-configuration.tsx
```diff
@@ -0,0 +1,315 @@
...
+};
+
+export function EmrConfiguration (props: EmrConfigurationProps) {
+    const applicationConfig: IAppConfiguration = useAppSelector(appConfig);
+    const emrApplicationList: string[] = useAppSelector(emrApplications);
+    const [selectedApplicationOptions, setSelectedApplicationOptions] = useState([] as any[]);
+    const [applicationOptions, setApplicationOptions] = useState([] as any);
+
+    useEffect(() => {
+        const appList: any[] = [];
```
dustins, line 61: Looks like we're adding objects like `{value: string, label: string}` below, can we use that instead of `any`?

frontend/src/entities/configuration/configuration-import-modal.tsx
```diff
@@ -0,0 +1,95 @@
...
+    FileUpload,
+    Header,
+    Modal,
+    SpaceBetween
+} from '@cloudscape-design/components';
+
+export type ConfigurationImportModalProps = {
+    visible: boolean;
+    selectedFile: File[];
+    setSelectedFile: (any) => void;
```
dustins, line 29: Is `setSelectedFile` only taking in `File[]` that we could use that instead of `any`?

frontend/src/entities/configuration/activated-services-configuration.tsx
```diff
@@ -0,0 +1,81 @@
...
+    enabledServices: any;
```
dustins, line 34: Looks like this is used as `{[key: string]: boolean}`?

frontend/src/entities/configuration/allowed-instance-types-configuration.tsx
```diff
@@ -0,0 +1,120 @@
...
+    ExpandableSection,
+    Header
+} from '@cloudscape-design/components';
+import React, { useMemo } from 'react';
+import { InstanceTypeMultiSelector } from '../../shared/metadata/instance-type-dropdown';
+
+export type AllowedInstanceTypesConfigurationProps = {
+    setFields: (any) => void;
+    expandedSections: {object};
+    setExpandedSections: (any) => void;
```
dustins, line 28: `expandedSections` looks like `{[key: string]: boolean}` might work and then for the set method too.

AlejandroRigau:
Adding here for posterity: This PR also includes changes from https://github.com/awslabs/mlspace/pull/123

frontend/src/entities/configuration/allowed-instance-types-configuration.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <Container
+                header={
+                    <Header variant='h2'
+                        description='Modify the Amazon EC2 instance types that you wish to make available to users within the following resources: Notebook instance, Training and HPO jobs, Transform jobs, and Endpoints.'
+                    >
+                        Allowed Instance Types
+                    </Header>
+                }
+            >
+                <ExpandableSection headerText='Notebook instances' variant='default' expanded={props.expandedSections.notebookInstances} onChange={({ detail }) =>
```
AlejandroRigau, line 79: Do these still make sense to have as expandable sections? Expanding and contracting only saves us the space of an input field which at that point, might as well always have it open.

frontend/src/entities/configuration/deployment-configuration.tsx
```diff
@@ -123,31 +122,34 @@ export function DeploymentConfiguration () {
             }
```
AlejandroRigau, line 122: Do we still need this header and top level container? I think it makes sense to have 
```
<>
    <Container></Container>
    <Container></Container>
</>
```
Rather than
```
<Container>
    <Container></Container>
    <Container></Container>
</Container>
```

This approach seems to match the other tabs

frontend/src/entities/configuration/project-creation-configuration.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <Container
+                header={
+                    <Header variant='h2'
+                        description='All users can create new projects in MLSpace. Restrict project creation to users with Admin permissions by activating “Admin Only.”'
+                    >
+                        Project Creation
+                    </Header>
+                }>
+                <Grid gridDefinition={[{ colspan: 2 }]}>
+                    <div style={{display: 'grid', textAlign: 'center'}}>
```
AlejandroRigau, line 42: I think you might want to use `<Box textAlign="center">` instead of inline CSS. Maybe with a `<SpaceBetween direction="vertical" size="xs">` 
https://cloudscape.design/components/box/
https://cloudscape.design/components/space-between/

frontend/src/entities/configuration/project-creation-configuration.tsx
```diff
@@ -0,0 +1,60 @@
...
+                        <div style={{display: 'grid', justifyContent: 'center'}}>
+                            <Toggle
```
AlejandroRigau, line 44: Why not use the label and description from the toggle? Just for visual purposes or is there another reason?
```
<Toggle
      onChange={({ detail }) =>
        rops.setFields({'configuration.ProjectCreation.isAdminOnly': detail.checked});
      }
      checked={props.isAdminOnly}
      description={
          Restrict project creation to users with Admin
          permissions
      }
    >
      Admin Only
    </Toggle>
```
Screenshot 2024-06-11 at 11 40 22 AM

frontend/src/entities/configuration/system-banner-configuration.tsx
```diff
@@ -0,0 +1,120 @@
...
+        <>
+            <Container
+                header={
+                    <Header variant='h2'>
+                        System Banner
+                    </Header>
+                }>
+                <SpaceBetween direction='vertical' size='l'>
+                    <Grid gridDefinition={[{colspan: 4}, {colspan: 4}, {colspan: 4}]}>
+                        <div style={{display: 'grid', textAlign: 'center'}}>
```
AlejandroRigau, line 48: Same as above. Maybe Box container can help avoid inline CSS.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -260,482 +187,110 @@ export function DynamicConfiguration () {
...
-
-            if (isObject(value)) {
-                const recursiveJson = jsonToOutline(value); // recursively call
-                output.push((recursiveJson));
-            }
-        }
-        return <ul>{output}</ul>;
-    }
-
     const changesDiff = getJsonDifference(applicationConfig.configuration, state.form.configuration);
```
AlejandroRigau, line 190: we should probably move this and the function to the confirm-configuration-changes modal file like you did with the  jsonToOutline file. they go hand in hand and I dont think changesDiff is use anywhere else other than that modal.

frontend/src/entities/configuration/project-creation-configuration.tsx
estohlmann: My task said to have the items it is toggling off below the toggle, this was the way to accomplish this. I can look into using the box, space-between components you called out. I didn't try with those yet. Not sure it will accomplish what we are looking for or not.

frontend/src/entities/configuration/allowed-instance-types-configuration.tsx
estohlmann: There were talks about removing some of these containers, the empty tags don't really add much so didn't think it was a big deal having them here. I will remove.

frontend/src/entities/configuration/allowed-instance-types-configuration.tsx
estohlmann: Some of these can be long lists, I can see the value in these being expanding. If we want to remove them for containers we can make a new task/PR to address as this wasn't called out in the initial task.

frontend/src/entities/configuration/deployment-configuration.tsx
estohlmann: I asked Cindy this when we were going through the text and she said she liked the current layout. All the tabs we have currently look identical.

frontend/src/entities/configuration/allowed-instance-types-configuration.tsx
dustins: It isn't a big deal, we've just been trying to not have them when not necessary.

frontend/src/entities/configuration/configuration-import-modal.tsx
estohlmann: Not sure, I removed it.

AlejandroRigau (approved):
Awesome work addressing all the comments. LGTM

douglas1850 (approved):

frontend/src/entities/configuration/allowed-instance-types-configuration.tsx
```diff
@@ -0,0 +1,118 @@
...
+import {
+    Container,
+    ExpandableSection,
+    Header
+} from '@cloudscape-design/components';
+import React, { useMemo } from 'react';
+import { InstanceTypeMultiSelector } from '../../shared/metadata/instance-type-dropdown';
+
+export type AllowedInstanceTypesConfigurationProps = {
+    setFields: (SetFieldsFunction) => void;
```
dustins, line 26: Not sure this PR is ready for review yet but pretty sure this should be `setFields: SetFieldsFunction`. I probably wasn't clear in my previous comment that `SetFieldsFunction` is the whole method signature instead of just the parameter signature.

frontend/src/entities/configuration/activated-services-configuration.tsx
dustins: Does this signature not work?

frontend/src/entities/configuration/configuration-history-table.tsx
```diff
@@ -36,7 +36,8 @@ export function ConfigurationHistoryTable () {
 
     useMemo(() => {
         dispatch(listConfigurations({configScope: 'global', numVersions: applicationConfig.versionId + 1}));
-    }, [applicationConfig.versionId, dispatch]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
dustins, line 39: What is the issue here that we need to disable exhaustive-deps?

frontend/src/entities/configuration/activated-services-configuration.tsx
estohlmann: I missed this, sorry :D

frontend/src/entities/configuration/configuration-history-table.tsx
estohlmann: No issue, will remove, I stole this from an issue I had someone fix while I was working on the refactor.

frontend/src/entities/configuration/activated-services-configuration.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const configurableServices = {
+    batchTranslate: 'Amazon Translate batch',
+    realtimeTranslate: 'Amazon Translate real-time',
+    emrCluster: 'Amazon EMR',
+    labelingJob: 'Amazon Ground Truth'
+};
+
+export type ActivatedServicesConfigurationProps = {
+    setFields: (values: { [key: string]: any }, method?: ValidationFormActionMethod) => void;
```
dustins, line 38: This is the right signature, and it makes no difference but I'd consider just importing `SetFieldsFunction` and having `setFields: SetFieldsFunction` so it is cleaner. But not 100% necessary.

frontend/src/entities/configuration/emr-configuration.tsx
dustins: Is this the wrong type or does this just not make sense to do?

dustins (approved):

PR #125, opened 2024-06-11
Add config history table pagination, filtering, and sorting

AlejandroRigau:
*Issue #, if available:* N/A

*Description of changes:*
- Add pagination, filtering, and sorting to config table

Recording 2024-06-11 at 12 26 14

### DIFF
--- a/frontend/src/entities/configuration/configuration-history-table.tsx
+++ b/frontend/src/entities/configuration/configuration-history-table.tsx
@@ -35,8 +35,8 @@ export function ConfigurationHistoryTable () {
     });
 
     useMemo(() => {
-        dispatch(listConfigurations({configScope: 'global', numVersions: 999999}));
-    }, [dispatch]);
+        dispatch(listConfigurations({configScope: 'global', numVersions: applicationConfig.versionId + 1}));
+    }, [applicationConfig.versionId, dispatch]);
 
     const columnDefinition = [
         {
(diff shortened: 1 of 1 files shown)


frontend/src/entities/configuration/configuration-history-table.tsx
```diff
@@ -34,20 +35,23 @@ export function ConfigurationHistoryTable () {
...
             cell: (item) => item.versionId,
         },
         {
             header: 'Changed by',
+            sortingField: 'changedBy',
             cell: (item) => item.changedBy,
         },
         {
             header: 'Change reason',
+            sortingField: 'changeReason',
```
dustins, line 54: I'm not sure it makes sense to sort by changeReason and possibly not changedBy too since there is a search field if you want to look at a specific person or reason. I guess it doesn't hurt to have but could be a little confusing to list things in other than numeric order.

dustins (dismissed):

frontend/src/entities/configuration/configuration-history-table.tsx
AlejandroRigau: by default it filters by version number but it cant hurt to allow users to filter by whatever they want

frontend/src/entities/configuration/configuration-history-table.tsx
```diff
@@ -34,20 +35,23 @@ export function ConfigurationHistoryTable () {
     });
 
     useMemo(() => {
-        dispatch(listConfigurations({configScope: 'global', numVersions: 10}));
+        dispatch(listConfigurations({configScope: 'global', numVersions: 999999}));
```
douglas1850, line 38: In reality I'm sure this works fine, but a more precise `numVersions` value would be passing in the versionId of the app config returned by `getConfiguration` (plus 1) since there will be up to `versionId + 1` versions.

douglas1850 (approved):

frontend/src/entities/configuration/configuration-history-table.tsx
AlejandroRigau: That's a significantly more clever solution. Even though I have all approvers, I want to change it to that. It makes more sense.

dustins (approved):

PR #128, opened 2024-06-12
Dynamic roles initial app config policy update

dustins:
*Issue #, if available:*

*Description of changes:*
On initial deploy with our new app config changes we need to update existing dynamic user roles. This addition the the initial app config lambda does just that. It also adds a `dynamic-user-role` tag to any dynamic user roles it finds. This tag is also going to be added in the future to any dynamic user roles we create so we can target them easier in the future.


backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -386,15 +386,20 @@ def _create_iam_policy(
...
         policy_version: int,
     ) -> str:
+        tags = [
+            {"Key": policy_type.lower(), "Value": policy_identifier},
+            {"Key": "policyVersion", "Value": str(policy_version)},
+            {"Key": "system", "Value": self.system_tag},
+        ]
+
+        if policy_type.lower() == "user":
+            tags.append({"Key": "dynamic-user-role", "Value": "true"})
```
LightSeekerSC, line 396: This is a policy, not a role. I think you meant to add this to `_create_iam_role`

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
```diff
@@ -44,5 +45,30 @@ def lambda_handler(event, context):
     ]
 
     app_configuration_dao.update(config)
+    update_dynamic_roles()
 
     generate_html_response(200, "Successfully updated app config")
+
+
+def update_dynamic_roles():
```
LightSeekerSC, line 53: This may work for your allowed instances policies, but will not work for allowed services policies as the policy may not be created yet. We probably need to make this an iam_manager function and then it can be called here and when creating the deny services policy

dustins: Can you expand on why this won't work with the service policies?

backend/src/ml_space_lambda/utils/iam_manager.py
dustins: You are absolutely right, great catch!

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
AlejandroRigau, line 53: Do we have to add some check to see if we are managing the IAM roles for them? I think we have the var `export const MANAGE_IAM_ROLES = true;`

lib/stacks/infra/core.ts
```diff
@@ -454,7 +457,7 @@ export class CoreStack extends Stack {
             role: props.mlSpaceAppRole
         });
 
-        new AwsCustomResource(this, 'populate-allowed-instance-types', {
+        new AwsCustomResource(this, 'initial-app-config-deployment-001', {
```
AlejandroRigau, line 460: nit: do we need the `001`?

dustins: This only ever runs once. So for us to get it to re-run we have to modify the name. Instead of coming up with unique names each time I figured a counter like this made more sense.

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
dustins: I just added that env var to the lambda and put in a check to only attempt this if `MANAGE_IAM_ROLES` is true.

LightSeekerSC: The services policy doesn't exist on deployment

LightSeekerSC: If we end up creating a group policy too then this will also be relevant there. We will want/need to apply this for dynamically created policies

dustins: I'll can update the name as this is a bit too generic but this is just for updating _existing_ dynamic user roles that were created **before** the instance type constraint policies existed on the notebook role. It should only run once the first time 1.5 is deployed.

Any dynamic roles created after 1.5 is deployed will be tagged appropriately and will have the correct notebook policies.

douglas1850 (approved):

AlejandroRigau (approved):

douglas1850 (approved):

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
LightSeekerSC: What I am saying is that we need this same logic for the deny role the first time it is created and for any new policies we may create in the future. This is reusable if you put it in the `iam_manager`. If you just want to commit this, then I can do the work to move it later, but I also need this applied to the service policy or otherwise customers will need to manually attach that.

backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -357,6 +357,9 @@ def remove_all_user_roles(self, username: str, projects: List[str]) -> None:
 
     # Creates the IAM role for the MLSpace user/project context
     def _create_iam_role(self, iam_role_name: str, project_name: str, username: str) -> str:
+        tags = generate_tags(username, project_name, self.system_tag)
+        tags.append({"Key": "dynamic-user-role", "Value": "true"})
```
LightSeekerSC, line 361: Doesn't it make more sense to add this to `generate_tags` as an optional named parameter?

I also feel like having value True is wasting opportunity here. Maybe `{ "role-type": "dynamic-user-role" }`?

I feel like this offers more opportunity for variety in the future.

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
LightSeekerSC: I still feel pretty strongly about moving this to the `iam_manager`, but if you don't do it, then I can do it in a separate PR. We need to update existing dynamic roles with the deny services policy too, so this is something that is needed elsewhere.

backend/src/ml_space_lambda/utils/iam_manager.py
dustins: Yeah I didn't love this. I can see accepting a list of additional tags to clean things up a bit.

I'm not as sure about changing the tag though. I get the instinct but since we don't have a use-case for other types of roles it seems potentially limiting (say we wanted roles to be multiple types in the future or something.) I don't think we really lose anything by having separate tags other than having the value be pretty useless.

AlejandroRigau (approved):

backend/src/ml_space_lambda/utils/iam_manager.py
LightSeekerSC: Even if you have hybrid in the future we can just adapt the type naming. Example:
- `dynamic-user-role`
- `dynamic-user-extra-role`

This boils down to preference I think, but I would lean away from creating a pattern that will bloat our tags.

dustins: I think I'm not going to do this for now (partially out of concern for for the time it would take to change it and the tests) but we can talk about this at another dev sync and always migrate in the same way we're adding these in the first place.

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
```diff
@@ -44,5 +46,24 @@ def lambda_handler(event, context):
...
+    if env_vars["MANAGE_IAM_ROLES"] == "false":
+        return
+
+    iam_manager = IAMManager(iam)
+    role_names = iam_manager.find_dynamic_user_roles()
+    policy_arns = [
+        env_vars["JOB_INSTANCE_CONSTRAINT_POLICY_ARN"],
+        env_vars["ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN"],
+    ]
+    iam_manager.attach_policies_to_dynamic_user_roles(policy_arns, role_names)
```
LightSeekerSC, line 66: I would do one of the following:
- Update the function name to `attach_policies_to_roles` OR
- Get the list of dynamic role names within the function

I don't currently anticipate any cases we would want to attach to any other roles, so I think option 2 of combining these makes more sense.

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
```diff
@@ -44,5 +46,24 @@ def lambda_handler(event, context):
...
     app_configuration_dao.update(config)
+    update_dynamic_roles_with_notebook_policies()
 
     generate_html_response(200, "Successfully updated app config")
+
+
+def update_dynamic_roles_with_notebook_policies():
+    env_vars = get_environment_variables()
+
+    if env_vars["MANAGE_IAM_ROLES"] == "false":
```
LightSeekerSC, line 57: I think you just want `if not env_vars["MANAGE_IAM_ROLES"]:` here. I think that strings are technically truthy. I think technically the value some users may use is ""

Everywhere else in the codebase we do `if env_variables["MANAGE_IAM_ROLES"]:`

dustins: Looks like for consistence I shouldn't be setting true/false in the environment variables but "true"/"". I updated the lambda definition so this usage matches the rest of the project.

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
dustins: I'm going to take your first suggestion and change it to `attach_policies_to_roles` because if you look below I'm using the `role_names` again. This is a relatively costly thing to query for so I dont' want to do that multiple times if not necessary. I thought giving the option to supply them so they arent' fetched dynamically would be good but I agree it is slightly weird.

douglas1850 (approved):

AlejandroRigau (approved):

LightSeekerSC (approved):

PR #129, opened 2024-06-12
restore high contrast header

AlejandroRigau:
*Issue #, if available:* N/A

Cloudscape made changes to change the default look of ContentLayouts: https://github.com/cloudscape-design/components/releases/tag/3.0.639

This PR adds the required settings to revert back to our original look. 

*Description of changes:*
- Create custom ContentLayout that sets high-contrast header
- Create horizontal line component to fix side navigation horizontal line

Screenshot 2024-06-12 at 2 36 38 PM

### DIFF
--- a/frontend/src/entities/configuration/configuration.tsx
+++ b/frontend/src/entities/configuration/configuration.tsx
@@ -66,13 +66,13 @@ export function Configuration () {
                             label: 'Dynamic Configuration',
                             id: 'dynamic-config',
                             content: <DynamicConfiguration/>,
-                            disabled: false
+                            disabled: true
                         },
                         {
                             label: 'Dynamic Configuration History',
                             id: 'config-hist',
                             content: <ConfigurationHistoryTable/>,
-                            disabled: false
+                            disabled: true
                         },
                         {
                             label: 'Deployment Configuration',
--- a/frontend/src/entities/configuration/configuration.tsx
+++ b/frontend/src/entities/configuration/configuration.tsx
@@ -19,7 +19,6 @@ import { setBreadcrumbs } from '../../shared/layout/navigation/navigation.reduce
 import { useAppDispatch } from '../../config/store';
 import {
     Header,
-    SpaceBetween,
     Tabs
 } from '@cloudscape-design/components';
 import React, { useEffect } from 'react';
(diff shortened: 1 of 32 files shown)


frontend/src/entities/configuration/configuration.tsx
```diff
@@ -66,13 +66,13 @@ export function Configuration () {
                             label: 'Dynamic Configuration',
                             id: 'dynamic-config',
                             content: <DynamicConfiguration/>,
-                            disabled: true
+                            disabled: false
```
estohlmann, line 69: We can just delete these, not sure we will ever need to have these again since it defaults to false.

AlejandroRigau: I actually didnt mean to include this change. I think @LightSeekerSC was going to add that in his PR

estohlmann: Agreed

AlejandroRigau:
Screenshot 2024-06-12 at 2 56 07 PM

Setting it back to draft. Seems that this change make the config page look off

frontend/src/shared/layout/horizontal-line.tsx
```diff
@@ -0,0 +1,32 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import { ColumnLayout } from '@cloudscape-design/components';
+import React from 'react';
+
+
+export const HorizontalLine = () => {
```
LightSeekerSC, line 21: Why are we adding this? This is kind of a weird way to effectively create a `<hr>` that we don't have any control over. I think creating an actual `<hr>` would be preferable, but I am confused why we need this in the first place.

PR looks good to me otherwise

AlejandroRigau: `<hr>` is a normal html horizontal line without styling while this component is a styled horizontal line that matches the Cloudscape UI design. Unfortunately Cloudscape does not have a HorizontalLine component that exposes their styling, but this is a hacky way to obtain the same result.

LightSeekerSC (approved):

dustins (approved):

PR #130, opened 2024-06-12
Creates unifed error handling for graceful handling of exceptions

LightSeekerSC:
*Description of changes:* Creates a unified function for handling error notifications, so that common exceptions can be identified and handled gracefully.

For discussion, it may be worthwhile to limit this specifically to certain error types (e.g. API errors) and to increase the input parameters to include the name of the resource for more personalized messages.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -62,9 +62,9 @@
             }
         },
         "node_modules/@adobe/css-tools": {
-            "version": "4.3.3",
-            "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.3.3.tgz",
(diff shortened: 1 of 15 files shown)


frontend/src/modules/modal/resource-schedule-modal.tsx
```diff
@@ -84,12 +84,12 @@ function ResourceScheduleModal ({
...
-                    success ? 'Successfully modified' : 'Failed to modify'
-                } auto-termination configuration for ${resourceName}.`,
-                success ? 'success' : 'error'
-            );
+            if (success) {
+                notificationService.generateNotification(
+                    `Successfully modified auto-termination configuration for ${resourceName}.`, 'success'
+                );
+            }
+            notificationService.handleErrorNotification(`Failed to modify auto-termination configuration for ${resourceName}.`);
```
dustins, line 92: Should this be an else instead of just coming after the if statement?

AlejandroRigau (changes requested):

LightSeekerSC:
Going to work off of Alejandro's suggestion and move this to the backend.

PR #131, opened 2024-06-12
App config documentation

douglas1850:
*Description of changes:* Updating the documentation to include information about the app config feature.

* info about the new roles (like the system role)
* info about the app config table being added
* instructions for how a customer can create their own system role and policy
* a number of whitespace and other formatting corrections to make the documentation more uniform 
* created a document for the configurations page, detailing what each section does


frontend/docs/admin-guide/getting-started.md
```diff
@@ -162,6 +174,18 @@ that project. The partition key is `resourceType` and the sort key is `resourcei
 | project | String | The project the resource is associated with. |
 | user | String | The user who created the resource. |
 
+#### Application Configuration Table
+
+The app config table contains dynamic configuration settings for {{ $params.APPLICATION_NAME }}. The table is populated with default values the first time it is deployed, and then admins can modify the dynamic configurations for {{ $params.APPLICATION_NAME }} by navigating to the `configuration` page accessible only to admins.
```
AlejandroRigau, line 179:
```suggestion
The app configuration table contains dynamic configuration settings for {{ $params.APPLICATION_NAME }}. The table is populated with default values the first time it is deployed, and then admins can modify the dynamic configurations for {{ $params.APPLICATION_NAME }} by navigating to the `Configuration` page accessible only to admins.
```

AlejandroRigau (commented):
LGTM so far.

README.md
```diff
@@ -172,12 +172,18 @@ If the config-helper doesn't provide the level of customization you need for you
 | EXISTING_VPC_NAME | If MLSpace is being deployed into an existing VPC this should be the name of that VPC (must also set `EXISTING_VPC_ID`) | - |
 | EXISTING_VPC_ID | If MLSpace is being deployed into an existing VPC this should be the id of that VPC (must also set `EXISTING_VPC_NAME`) | - |
 | EXISTING_VPC_DEFAULT_SECURITY_GROUP | If MLSpace is being deployed into an existing VPC this should be the default security group of that VPC | - |
+<<<<<<< HEAD
```
AlejandroRigau, line 175: Probably left over form merge

README.md
```diff
@@ -172,12 +172,18 @@ If the config-helper doesn't provide the level of customization you need for you
...
 | SYSTEM_ROLE_ARN | Arn of an existing IAM role to use for executing the MLSpace system lambdas. System lambdas are responsible for maintaining the MLSpace system by cleaning up resources when a project is suspended or deleted, when a user is suspended, or when services are activated/deactivated. This value must be set to an existing role because the default CDK deployment will not create one. | - |
 | ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN | ARN for policy constraining the instance size that can be used when creating Endpoint configurations from a notebook. | - |
 | JOB_INSTANCE_CONSTRAINT_POLICY_ARN | ARN for policy constraining the instance size that can be used when creating HPO/Training/Transform jobs from a notebook. | - |
 | S3_READER_ROLE_ARN | Arn of an existing IAM role to use for reading from the static website S3 bucket. If not specified a new role with the correct privileges will be created | - |
+=======
```
AlejandroRigau, line 182: Probably left over form merge

README.md
```diff
@@ -172,12 +172,18 @@ If the config-helper doesn't provide the level of customization you need for you
...
+| APP_ROLE_ARN | ARN of an existing IAM role to use for executing the MLSpace lambdas. This value must be set to an existing role because the default CDK deployment will not create one. | - |
+| NOTEBOOK_ROLE_ARN | ARN of an existing IAM role to associate with all notebooks created in MLSpace. If using dynamic roles based on project/user combinations the specific combination role will be used instead. This value must be set to an existing role because the default CDK deployment will not create one. | - |
+| S3_READER_ROLE_ARN | ARN of an existing IAM role to use for reading from the static website S3 bucket. If not specified a new role with the correct privileges will be created | - |
+>>>>>>> 9f37c44b84e0211f9832e0239e99d4ec2e52e2dc
```
AlejandroRigau, line 186: Probably left over form merge

README.md
```diff
@@ -172,12 +172,18 @@ If the config-helper doesn't provide the level of customization you need for you
...
 | SYSTEM_ROLE_ARN | Arn of an existing IAM role to use for executing the MLSpace system lambdas. System lambdas are responsible for maintaining the MLSpace system by cleaning up resources when a project is suspended or deleted, when a user is suspended, or when services are activated/deactivated. This value must be set to an existing role because the default CDK deployment will not create one. | - |
 | ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN | ARN for policy constraining the instance size that can be used when creating Endpoint configurations from a notebook. | - |
 | JOB_INSTANCE_CONSTRAINT_POLICY_ARN | ARN for policy constraining the instance size that can be used when creating HPO/Training/Transform jobs from a notebook. | - |
 | S3_READER_ROLE_ARN | Arn of an existing IAM role to use for reading from the static website S3 bucket. If not specified a new role with the correct privileges will be created | - |
+=======
+| APP_ROLE_ARN | ARN of an existing IAM role to use for executing the MLSpace lambdas. This value must be set to an existing role because the default CDK deployment will not create one. | - |
```
AlejandroRigau, line 183: APP_ROLE_ARN, S3_READER_ROLE_ARN and NOTEBOOK_ROLE_ARN seem to also be set further up.

dustins (dismissed):

README.md
douglas1850: Ahh thanks, I thought I completed the merge but I guess this snuck through

dustins (approved):

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
 13. After selecting the 3 policies click next and enter a name for the role. You can name the role whatever you'd like. Optionally add a description and tags and then click "Create role"
 14. Once the role has been created record the role ARN as we'll need to use it later.
 
+#### System Role
+
+1. The example below assumes you've created the permission boundary policy above in the [App Role](App Role) section and named the policy `mlspace-project-user-permission-boundary`. If you've named it something different
+you'll need to update the `"iam:PermissionsBoundary` condition in the `iam:CreateRole` statement below.
+2. After the permission boundary policy has been created you are now ready to create the application
+policy. From the IAM Service page click "Policies" on the left hand side.
+3. Create a new policy using the JSON editor and paste the following in the text area (after replacing the placeholder variables):
```
LightSeekerSC, line 978: It may be worth noting that they can use the app and notebook policies or if they want to restrict to least privileges they can use this one (saves them creating a new policy if they want to do that)

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
 13. After selecting the 3 policies click next and enter a name for the role. You can name the role whatever you'd like. Optionally add a description and tags and then click "Create role"
 14. Once the role has been created record the role ARN as we'll need to use it later.
 
+#### System Role
+
+1. The example below assumes you've created the permission boundary policy above in the [App Role](App Role) section and named the policy `mlspace-project-user-permission-boundary`. If you've named it something different
+you'll need to update the `"iam:PermissionsBoundary` condition in the `iam:CreateRole` statement below.
+2. After the permission boundary policy has been created you are now ready to create the application
```
LightSeekerSC, line 976: I think this should say "system policy", but also see my note on that they can re-use the app policy if they desire.

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+                "s3:Get*",
+                "s3:DeleteObject",
+                "s3:PutBucketNotification"
+            ],
+            "Resource": "arn:{AWS_PARTITION}:s3:::*",
+            "Effect": "Allow"
+        },
+        {
+            "Action": [
+                "sagemaker:UpdateNotebookInstanceLifecycleConfig"
```
LightSeekerSC, line 1047: Probably also need describe and delete here
```
'sagemaker:DeleteNotebookInstanceLifecycleConfig',
'sagemaker:DescribeNotebookInstanceLifecycleConfig',
```

Looks like they are in a separate block below. "Resource" and "Effect" seem to be the same with no condition, so I think they can be consolidated.

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+            "Condition": {
+                "Null": {
+                    "aws:ResourceTag/project": "false",
+                    "aws:ResourceTag/system": "false",
+                    "aws:ResourceTag/user": "false"
+                }
+            },
+            "Action": [
+                "sagemaker:DeleteNotebookInstance",
+                "sagemaker:DescribeNotebookInstance",
```
LightSeekerSC, line 1062: Probably want 'sagemaker:UpdateNotebookInstance' update here just in case

AlejandroRigau (dismissed):

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+                "elasticmapreduce:DescribeCluster",
+                "elasticmapreduce:ListInstances",
+                "elasticmapreduce:TerminateJobFlows",
+                "elasticmapreduce:SetTerminationProtection"
+            ],
+            "Resource": "arn:{AWS_PARTITION}:elasticmapreduce:{AWS_REGION}:{AWS_ACCOUNT}:cluster/*",
+            "Effect": "Allow"
+        },
+        {
+            "Action": [
```
LightSeekerSC, line 1107: So I take it the goal here is that they don't attach the notebook policy either? If you are planning on doing a single policy, then you will also need all the stop, list, and describe permissions from our notebook policy added in this policy.

I like the idea of giving the customer the option to choose convenience of reusing the notebook and app policy or using least-privileges security by creating this role. Additionally, if they choose to re-use app policy and notebook policy it reduces need for duplicate updates in the future if they are needed.

TLDR:
- If reusing notebook policy remove this
- If consolidating notebook policy permissions, I think permissions related to other services are missing

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+            "Effect": "Allow"
+        },
+        {
+            "Condition": {
+                "StringEqualsIgnoreCase": {
+                    "iam:ResourceTag/system": "MLSpace"
+                }
+            },
+            "Action": [
+                "iam:DetachRolePolicy",
```
LightSeekerSC, line 1122: If the update configuration lambda is going to use this, then we have to add attach here:
```
'iam:AttachRolePolicy'
```

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+                "iam:DeleteRole",
+                "iam:DeleteRolePolicy",
+            ],
+            "Resource": "arn:{AWS_PARTITION}:iam::{AWS_ACCOUNT}:role/MLSpace*",
+            "Effect": "Allow"
+        },
+        {
+            "Action": [
+                "iam:ListRoles",
```
LightSeekerSC, line 1131: I guess we can have the project create and user create policy use the App role, but it feels like a weird line to not include creating roles here and to split the project and user lambdas between different roles. 

It feels like creating roles is something the System role should be able to do.

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+                "iam:ListAttachedRolePolicies",
+                "iam:GetRole",
+                "iam:GetPolicy"
+            ],
+            "Resource": "*",
+            "Effect": "Allow"
+        },
+        {
+            "Action": [
+                "iam:DeletePolicy",
```
LightSeekerSC, line 1143: The app config update lambda needs the following permissions:
```
'iam:CreatePolicy',
'iam:CreatePolicyVersion',
'iam:TagPolicy',
```

It needs to be able to modify dynamic roles. Again this is something that other lambdas working with dynamic roles will need and something I think should be standard for the System role

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+            "Action": "iam:PassRole",
+            "Resource": "arn:{AWS_PARTITION}:iam::{AWS_ACCOUNT}:role/{MLSPACE_APP_ROLE_NAME}",
+            "Effect": "Allow"
+        }
+    ]
+}
+```
+
+4. Click next and optionally add tags to this policy
+5. Click next again and enter a name for this policy. You can name the policy whatever you'd like (ex: systemPolicy) but ensure you remember it as you'll need it when creating the role.
```
LightSeekerSC, line 1174: I would change the recommended name to have an `mlspace-` prefix unless we are intentionally recommending this to avoid collisions.

frontend/docs/admin-guide/install.md
```diff
@@ -969,6 +969,232 @@ policy. From the IAM Service page click "Policies" on the left hand side.
...
+            "Principal": {
+                "Service": "lambda.amazonaws.com"
+            },
+            "Action": "sts:AssumeRole"
+        }
+    ]
+}
+```
+
+9. Click the next button and then select the checkbox next to the name of the policy you created in step 5 above. You will also need to attach the default notebook policy you previously created as well as the AWS managed policy `AWSLambdaVPCAccessExecutionRole`. In total you should have 3 policies attached to the role.
```
LightSeekerSC, line 1194: Here notes attaching the notebook policy. This does have create permissions included in it. Up to you how you want to handle that, but if they are going least privileges, then they really should consolidate all the permissions in the other policy above.

frontend/docs/user-guide/administration-pages.md
```diff
@@ -17,13 +17,12 @@ associated with the suspended user.
 
 ## Configuration
 
-Administrators can view the current configurations for {{ $params.APPLICATION_NAME }}. If configuration settings need to be changed,
-a system admin developer will need to make those modifications.
+See the [configuration page documentation](./configuration.md).
```
LightSeekerSC, line 20: I would provide at least a little bit of detail here and then a "For more information, see the configuration page documentation."

They should be able to still get a high-level overview from this page I think.

AlejandroRigau (approved):

frontend/docs/user-guide/configuration.md
```diff
@@ -0,0 +1,63 @@
+---
+outline: deep
+---
+
+# Configuration
+
+Users with Administrator permission can access {{ $params.APPLICATION_NAME }}’s Configuration page. This page is divided into three sections:
+
+1. Dynamic Configuration
```
LightSeekerSC, line 9: I think we could add [MD heading links here like a table of contents](https://www.markdownguide.org/extended-syntax/#linking-to-heading-ids)

frontend/docs/user-guide/configuration.md
```diff
@@ -0,0 +1,63 @@
...
+
+At the bottom of this page is a "Save Changes" button. Upon clicking this, the user will be shown a brief summary of the fields that have been changed and will be updated when this new dynamic configuration is uploaded.
+
+### Allowed Instance Types
+
+Modify the Amazon EC2 instance types that you wish to make available to users within the following resources: Notebook instance, Training and HPO jobs, Transform jobs, and Endpoints. IAM permissions that control access to these instance types within the {{ $params.APPLICATION_NAME }} user interface and Jupyter Notebooks will automatically update.
+
+### Activated Services
+
+Activate or deactivate services within {{ $params.APPLICATION_NAME }}. IAM permissions that control access to these services within the {{ $params.APPLICATION_NAME }} user interface and Jupyter Notebooks will automatically update. Deactivated services will no longer appear within the {{ $params.APPLICATION_NAME }} user interface. Deactivating services will terminate all active corresponding jobs and instances associated with the service.
```
LightSeekerSC: "terminate" -> "suspend"

Since this is the more detailed page, we may want to note that it will actually terminate EMR clusters because they are not able to be suspended

README.md
```diff
@@ -163,17 +164,20 @@ If the config-helper doesn't provide the level of customization you need for you
...
+| APP_ROLE_ARN | Arn of an existing IAM role to use for executing the MLSpace lambdas. This value must be set to an existing role because the default CDK deployment will not create one. | - |
+| NOTEBOOK_ROLE_ARN | Arn of an existing IAM role to associate with all notebooks created in MLSpace. If using dynamic roles based on project/user combinations the specific combination role will be used instead. This value must be set to an existing role because the default CDK deployment will not create one. | - |
+| SYSTEM_ROLE_ARN | Arn of an existing IAM role to use for executing the MLSpace system lambdas. System lambdas are responsible for maintaining the MLSpace system by cleaning up resources when a project is suspended or deleted, when a user is suspended, or when services are activated/deactivated. This value must be set to an existing role because the default CDK deployment will not create one. | - |
+| ENDPOINT_CONFIG_INSTANCE_CONSTRAINT_POLICY_ARN | ARN for policy constraining the instance size that can be used when creating Endpoint configurations from a notebook. | - |
```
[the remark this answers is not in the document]
douglas1850: @dustins correct me if I wrong, but it was worded like this because the authorizer + the UI simply not presenting these instance types will handle the UI usecase, whereas these IAM policies are here to stop notebook users from using these services right?

README.md
```diff
@@ -163,17 +164,20 @@ If the config-helper doesn't provide the level of customization you need for you
...
+| JOB_INSTANCE_CONSTRAINT_POLICY_ARN | ARN for policy constraining the instance size that can be used when creating HPO/Training/Transform jobs from a notebook. | - |
```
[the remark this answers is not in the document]
douglas1850: We'll see if what Dustin says, but same logic here

bin/scripts/config-helper.mjs
```diff
@@ -186,6 +186,11 @@ async function askRoleQuestions () {
             name: 'APP_ROLE_ARN',
             message: 'App Role ARN: arn of an existing IAM role to use for executing the MLSpace lambdas. This value must be set to an existing role because the default CDK deployment will not create one',
         },
+        {
+            type: 'input',
+            name: 'SYSTEM_ROLE_ARN',
+            message: 'System Role ARN: arn of an existing IAM role to use for executing the MLSpace system lambdas. This value must be set to an existing role because the default CDK deployment will not create one.',
```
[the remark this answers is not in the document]
douglas1850: Honestly I didn't understand that description either - it's basically copied from the Application role description.

I'm guessing that the baked-in assumption is that our customers will always need to create their own roles? But even then this description is not strictly true either.

I'll be interested to see if anyone has a good justification for that line in the application role, because if they don't I say we remove it from both descriptions.

frontend/docs/admin-guide/install.md
douglas1850: No they'll need the permission boundary policy like they have for the app role to ensure the system policy/role never has permissions outside the permission boundary.

frontend/docs/admin-guide/install.md
douglas1850: Sure I'll make a note of that

frontend/docs/admin-guide/install.md
douglas1850: That's in another section for this policy, I'll combine them (not sure why they were separate in the app policy I originally copied)

frontend/docs/admin-guide/install.md
douglas1850: It's out of scope for the documentation PR, but I'm open to the discussion of moving all permissions for project/user processes to the system role (and anything else that we think should we handled by a system role rather than an application role used by users).

frontend/docs/admin-guide/install.md
douglas1850: Good call out thanks

frontend/docs/admin-guide/install.md
douglas1850: Ehh I don't think it's something to worry about, in the other instructions (like for app role) we don't even give an example name. As long as the customer remembers what it is for the following steps, it doesn't really matter. Plus depending on the customer, this may be white-labeled and not be mlspace.

frontend/docs/admin-guide/install.md
douglas1850: Went through and added all of the notebook policies that were missing in this policy. Excluded stuff that obviously wasn't applicable. Also updated the instructions to have a caveat that you could either create this new policy, or re-use the app policy + notebook policy.

README.md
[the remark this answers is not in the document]
dustins: That's correct. Those policies are specific to notebooks.

But thinking about it I guess _technically_ a user could submit a job with the incorrect instance type if they bypassed the UI right? Is there any downside to attaching these to the app role too? (other than using 2 more policy slots on that role)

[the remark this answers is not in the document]
dustins: I think we should move forward with merging these changes, if I make changes to apply these to the app role I'll fix the docs.

dustins (approved):

README.md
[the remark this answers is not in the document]
douglas1850: Using up 2 more slots is potentially a significant drawback if we ever need those slots in the future. How would someone bypass the UI without using a notebook? You're thinking they could send a curl command in the browser terminal or something? 

At any rate these descriptions are indeed accurate as they're written today, so I'll resolve this thread.

frontend/docs/user-guide/configuration.md
douglas1850: Terminating a job is the same as stopping it, right? I think the wording is pretty accurate since disabling a service stops the jobs and terminates the EMR cluster (meaning the instances it uses).

bin/scripts/config-helper.mjs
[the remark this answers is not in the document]
LightSeekerSC: Maybe ask the PM then? I am assuming they helped with the original documentation or know why we have it this way.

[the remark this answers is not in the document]
douglas1850: She's out today so I'm not going to block on this; we can ask on Monday

README.md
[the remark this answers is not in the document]
LightSeekerSC: There is a semi-easy hack for this via the UI then. With dev tools I think you could modify a value to whatever instance you want and launch it since we aren't doing any validation against the list of allowed instances on the front or backend.

Not sure anyone would do this, but I think it is possible.

I would ask the PM about this, because if I was a customer this would be confusing. It does limit the UI as well, so saying it just affects notebooks still feels misleading.

frontend/docs/admin-guide/install.md
LightSeekerSC: I think you may have misunderstood me here. This is the "System Role" section and in this bullet it says, "you are now ready to create the **application** policy".

Shouldn't this say "system policy"?

frontend/docs/admin-guide/install.md
LightSeekerSC: Final sanity check, but can you try creating a policy with all of the combined statements as a sanity check that this is a valid length.

frontend/docs/user-guide/configuration.md
LightSeekerSC: Yes and no. Really the only service where this is a truly strong distinction is notebooks. Notebooks can both be stopped and terminated. However, technically, this does just suspend the resources and is now the terminology we use on the admin page, so it is probably better to be consistent.

README.md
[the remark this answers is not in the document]
douglas1850: Discussed in meeting; will leave the description as-is since it's currently true, but we're planning to update this so this policy is in the app role as well. We will update the documentation accordingly when the application code is udpated.

dustins (dismissed):

dustins (approved):

LightSeekerSC (approved):

PR #133, opened 2024-06-13
Hides dynamic config options only for dynamic roles

LightSeekerSC:
*Description of changes:* Hides instance and service customization if dynamic roles aren't being used due to the fact that these options won't operate as expected in controlling notebook usage without dynamic roles.


frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -239,6 +240,11 @@ export function DynamicConfiguration () {
                         text={state.form.configuration.SystemBanner.text}
                         backgroundColor={state.form.configuration.SystemBanner.backgroundColor}
                         textColor={state.form.configuration.SystemBanner.textColor} />
+                    { !window.env.MANAGE_IAM_ROLES && 
+                        <Alert statusIconAriaLabel='Info'>
+                            {window.env.APPLICATION_NAME} has the ability to configure which services are available and what instance types are available if dynamic roles are enabled
+                        </Alert>
+                    }
```
dustins, line 247: We should try and get wording from Cindy if we want to promote the usage of dynamic roles I think it could be a warning and make it obvious from the start. Maybe something like this?

image

estohlmann: Probably makes the most sense to have this at the top of the config page vs the bottom too, that's generally where these types of messages appear.

frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -201,17 +202,17 @@ export function DynamicConfiguration () {
                 }
             >
                 <SpaceBetween direction='vertical' size='xl'>
-                    <AllowedInstanceTypesConfiguration
+                    { window.env.MANAGE_IAM_ROLES && <AllowedInstanceTypesConfiguration
```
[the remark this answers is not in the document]
LightSeekerSC: From memory we have intentionally been doing the opposite of this request, which was why I chose this. Probably best to just agree as a team. I will send a message out to the team to solicit feedback

frontend/src/entities/configuration/dynamic-configuration.tsx
LightSeekerSC: I can move it. I will consult Cindy from a user perspective. My thought was that I wouldn't constantly have to have a semi non-relevant box displayed, but maybe that is the point.

douglas1850: >a little too judgemental 

but that's the point 😉

Top of the page has my vote as well - this admin page is something that will be used very intermittently. This way when they use the page again after not touching it for a few weeks/months and can't remember why 2 configurations aren't visible (which are in documentation) they'll know why.

LightSeekerSC:
Update based on feedback and desired verbiage from the PM

Screenshot 2024-06-13 at 4 41 14 PM

frontend/src/entities/configuration/dynamic-configuration.tsx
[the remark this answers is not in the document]
LightSeekerSC: Team discussion favored the `{}` format for the time being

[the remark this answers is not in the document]
AlejandroRigau: Seems that the team prefers this approach. Ignore my comment.

AlejandroRigau (approved):

dustins (approved):

estohlmann (approved):

PR #135, opened 2024-06-13
adding warning to save confirmation if user is disabling a service

estohlmann:
Adding a warning to the save confirmation modal if a user is disabling a service letting them know what will happen.

Screenshot 2024-06-13 at 11 04 36 AM

### DIFF
--- a/frontend/src/entities/configuration/confirm-configuration-changes-modal.tsx
+++ b/frontend/src/entities/configuration/confirm-configuration-changes-modal.tsx
@@ -139,9 +139,10 @@ export function ConfirmConfigurationChangesModal (props: ConfirmConfigurationCha
         >
             <SpaceBetween size={'s'}>
                 <Alert statusIconAriaLabel='Warning' visible={isDeactivatingService} type='warning'>
-                    You are about to deactivate a service: Deactivated services will no longer appear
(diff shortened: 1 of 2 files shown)


frontend/src/entities/configuration/confirm-configuration-changes-modal.tsx
```diff
@@ -133,6 +138,11 @@ export function ConfirmConfigurationChangesModal (props: ConfirmConfigurationCha
             }
         >
             <SpaceBetween size={'s'}>
+                <Alert statusIconAriaLabel='Warning' visible={isDeactivatingService} type='warning'>
+                    You are about to deactivate a service: Deactivated services will no longer appear
+                    within the MLSpace user interface. Deactivating services will terminate all active
```
LightSeekerSC, line 143: Should say "suspend" instead of "terminate". It does terminate EMR though, so not sure if that is worth noting (not possible to suspend EMR)

estohlmann: If that is the case should we update the call out in the activate services page?

dustins (approved):

LightSeekerSC (approved):

PR #141, opened 2024-06-14
Bugfix/system role name

douglas1850:
*Description of changes:* Even reverting the role name doesn't resolve the CDK deploy problem, so I'm going to just proceed with the role being named mlspace-system-role so it matches the formatting of other roles, and I'll clean up the CDK deployment myself.


dustins (approved):

estohlmann (approved):

estohlmann:
We are going to have to figure out what this means for 1.4 -> 1.5

AlejandroRigau (approved):

douglas1850:
For customers deploying 1.4 -> 1.5 for the first time this won't be an issue.

The problem we're having is that we deployed with an IAM role named "A", and then tried to change the name to "B". IAM roles can't change their names, so CDK is trying to delete the old one to swap it with the new one... but the role "A" is in-use, so it fails.

A new deployment won't have the old name, so there won't be any issues like this one.

In this case I can resolve this problem by manually deleting the project and app-config APIs and then deploying these changes. The project/app-config APIs are the only things using this role.

PR #143, opened 2024-06-14
App role name env fix

LightSeekerSC:
*Description of changes:* Fixes error where the name of the role being used for the lambda is inserted in the env "APP_ROLE_NAME" instead of the actual app role name


lib/stacks/api/datasets.ts
```diff
@@ -122,6 +122,7 @@ export class DatasetsApiStack extends Stack {
                 restApi,
                 props.authorizer,
                 props.applicationRole,
```
AlejandroRigau, line 124: did you mean to replace this one?

LightSeekerSC: Not sure what you mean, but this does look right.

estohlmann (approved):

estohlmann (approved):

lib/utils/apiFunction.ts
```diff
@@ -44,7 +44,8 @@ export function registerAPIEndpoint (
     stack: Stack,
     api: IRestApi,
     authorizer: IAuthorizer,
-    appRole: IRole,
+    role: IRole,
+    appRoleName: string,
```
dustins, line 48: I won't block on this but it feels a little unnecessary. We already have MLSpaceConfig being passed in, which has `APP_ROLE_ARN` on it and the name can be extracted from that right? Seems like a lot of changes to pass something in a different way when we could just derive it with what we have.

lib/stacks/api/datasets.ts
dustins: Admittedly I was confused at first too. The first of these two parameters looks to be a real `IRole` (sometimes it is the application role but sometimes not) and then the second is a string of the role name.

dustins (approved):

AlejandroRigau (approved):

lib/utils/apiFunction.ts
douglas1850: If this is true I feel like it's worth blocking on... no reason to add more complexity and make this harder for the guy person to understand/maintain.

LightSeekerSC: `APP_ROLE_ARN` isn't guaranteed. It is only added if the customer is manually managing IAM roles

lib/stacks/api/appConfiguration.ts
```diff
@@ -68,6 +68,7 @@ export class AppConfigurationApiStack extends Stack {
                 restApi,
                 props.authorizer,
                 system_permissions.includes(f.name) ? props.systemRole : props.applicationRole,
```
douglas1850, line 70: Same thing here - we sometimes use the system role, is that going to create a problem?

douglas1850:
My biggest confusion about this change is that in several places we're passing in:

```
...
                props.applicationRole,
                props.applicationRole.roleName,
...
```

So we're already passing in the role object, and then we're passing in the role object's roleName property as well - why can't we just use the roleName property of the Role object *inside* the function, why do we need an additional parameter?

lib/stacks/api/projects.ts
```diff
@@ -226,6 +226,7 @@ export class ProjectsApiStack extends Stack {
                 restApi,
                 props.authorizer,
                 system_permissions.includes(f.name) ? props.systemRole : props.applicationRole,
+                props.applicationRole.roleName,
```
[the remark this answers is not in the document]
douglas1850: Talked on slack; this is actually the entire problem. The update_configuration lambda uses the system role for execution, but needs the app role name to update it on config changes.

douglas1850 (approved):

PR #147, opened 2024-06-17
Deny services policy retroactive attachment

LightSeekerSC:
*Description of changes:* Moves dynamic role attachment within a conditional. Should not affect current processing, but prevents unintentional attachment to dynamic roles in the future.


dustins (dismissed):

douglas1850:
Can you just explain some more about what this is accomplishing and why we need it? Is this fixing a bug, an improvement?

LightSeekerSC:
This makes it so that we are not always attaching a policy to all previous dynamic roles on a create (current functionality). The only time we are calling "update_dynamic_policy" right now is a case where we would want this, but this prevents doing this in a case where it would not be wanted. Attaching to all dynamic roles should not be a default and should only be done when it is intentionally desired.

Looking into the tests.

dustins (approved):

AlejandroRigau (approved):

PR #148, opened 2024-06-17
enforce system banner text if banner is activated

AlejandroRigau:
*Issue #, if available:* N/A

*Description of changes:*
- enforce system banner text if banner is activated
- Stylistic changes to to beautify it

Recording 2024-06-17 at 14 16 42

### DIFF
--- a/frontend/src/entities/configuration/dynamic-configuration.tsx
+++ b/frontend/src/entities/configuration/dynamic-configuration.tsx
@@ -14,29 +14,29 @@
   limitations under the License.
 */
 import {
-    Header,
-    SpaceBetween,
-    Container,
+    Alert,
     Button,
     ButtonDropdown,
+    Container,
     Grid,
-    Alert
+    Header,
+    SpaceBetween
 } from '@cloudscape-design/components';
 import React, { useEffect, useState } from 'react';
+import { z } from 'zod';
 import { useAppDispatch, useAppSelector } from '../../config/store';
-import { appConfig, getConfiguration, updateConfiguration } from './configuration-reducer';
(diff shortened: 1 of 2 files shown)


frontend/src/entities/configuration/dynamic-configuration.tsx
```diff
@@ -259,11 +267,16 @@ export function DynamicConfiguration () {
...
+                                    setState({ validateAll: true, formSubmitting: false });
+                                    scrollToInvalid();
+                                } else {
+                                    setModalVisible(true);
+                                }
                             }}
                             loading={state.formSubmitting}
                             data-cy='dynamic-configuration-submit'
-                            disabled={!isValid || state.formSubmitting}
+                            disabled={state.formSubmitting}
```
douglas1850, line 279: Any reason you removed the `isValid` check here? Looking at your gif it looks like when the text field is blank, the "Save Changes" button is active.

AlejandroRigau: thats how we do it in all other form submits (ex: notebook create). If I don't do it this way, then the user would not know why the form is not valid unless we have some sort of useEffect running with setting ValidateAll. We need that validateAll call to trigger the "error" state of the formField.

Anyways, I prefer to keep it similar to what we do with all other forms unless there is a specific reason to have it different here

douglas1850: Fair enough, this is how we do it in many places

douglas1850 (approved):

dustins (approved):

PR #149, opened 2024-06-17
Bugfix/only fetch app config once

douglas1850:
*Description of changes:* When fetching the app config, we were seeing the `NS_BINDING_ABORTED` error in the network tab. This error occurs when we [make multiple requests to the same endpoint in a short amount of time](https://stackoverflow.com/questions/704561/ns-binding-aborted-shown-in-firefox-with-httpfox).

Basically what was happening is:
* web app loads, useMemo renders and request for app config is made
* useMemo is re-rendered because of dev mode *and* our `currentUser` dependency
* the first request is canceled because of the second request. The first request is thusly marked `NS_BINDING_ABORTED`
* Thunk, being the amazing tool that it is capable of no wrong, interprets this as a rejection, so we hit our `isRejection` logic which sets `failedToLoadConfig` to True and results in the error notification being rendered
* the second GET app-config call succeeds and populates the web app with the correct app config

These changes will ensure that:
1. no unnecessary re-renders occur
2. if any re-renders _do_ occur, we will skip the fetch app-config logic

Attached screenshot demonstrates that GET app-config is only called once (meanwhile you can see calls to things like current-user are made multiple times).

image


frontend/src/routes.tsx
```diff
@@ -77,16 +77,25 @@ export default function AppRoutes () {
     ]);
 
     useMemo(() => {
-        dispatch(getConfiguration({configScope: 'global'}));
+        let configFetched = false;
+
+        async function fetchAppConfig () {
+            if (!configFetched) {
```
LightSeekerSC, line 83: Won't this always be false since it is reset when useMemo() is called?

frontend/src/routes.tsx
```diff
@@ -77,16 +77,25 @@ export default function AppRoutes () {
...
+                await dispatch(getConfiguration({configScope: 'global'}));
+            }
+            configFetched = true;
+        }
+
+        fetchAppConfig();
```
AlejandroRigau, line 89: cant this be summarized to 
```suggestion
useMemo(async () => {
      await dispatch(getConfiguration({configScope: 'global'}));
```

frontend/src/routes.tsx
douglas1850: It's some React magic.

[Here's the documentation](https://react.dev/learn/synchronizing-with-effects#fetching-data), the relevant section is:
>In development, you will see two fetches in the Network tab. There is nothing wrong with that. With the approach above, the first Effect will immediately get cleaned up so its copy of the ignore variable will be set to true. So even though there is an extra request, it won’t affect the state thanks to the if (!ignore) check.

frontend/src/routes.tsx
douglas1850: The [React documentation](https://react.dev/learn/synchronizing-with-effects#fetching-data) makes it a point to include the boolean to keep track of the request has already been made. I could probably get away with:

```
    useMemo(async () => {
        let configFetched = false;
        if (!configFetched) {
            await dispatch(getConfiguration({configScope: 'global'}));
        }
        configFetched = true;
        if (configLoadError) {
            notificationService.generateNotification(
                'Error loading app configuration. Restrictive default policy has been applied in its place. Consult with system admin to resolve issue.',
                'error'
            );
        }
    // eslint-disable-next-line react-hooks/exhaustive-deps
    }, [
        configLoadError,
        dispatch,
    ]);
```

AlejandroRigau (dismissed):

frontend/src/routes.tsx
douglas1850: To really match their example, I'm currently testing:
```
    useMemo(async () => {
        let configFetched = false;
        if (!configFetched) {
            await dispatch(getConfiguration({configScope: 'global'}));
        }
        
        if (configLoadError) {
            notificationService.generateNotification(
                'Error loading app configuration. Restrictive default policy has been applied in its place. Consult with system admin to resolve issue.',
                'error'
            );
        }
        return () => {
            configFetched = true;
        };
    // eslint-disable-next-line react-hooks/exhaustive-deps
    }, [
        configLoadError,
        dispatch,
    ]);
```

frontend/src/routes.tsx
```diff
@@ -76,17 +76,23 @@ export default function AppRoutes () {
...
+            await dispatch(getConfiguration({configScope: 'global'}));
+        }
+        
         if (configLoadError) {
             notificationService.generateNotification(
                 'Error loading app configuration. Restrictive default policy has been applied in its place. Consult with system admin to resolve issue.',
                 'error'
             );
         }
+        return () => {
```
estohlmann, line 91: I don't think you need to return anything since you aren't setting this to a value.

estohlmann (approved):

dustins (approved):

AlejandroRigau (approved):

PR #153, opened 2024-06-18
Docs: Fix broken headers containing variables

dustins:
*Issue #, if available:*

*Description of changes:*
Reverting update that changed sections to headers. This breaks the search functionality so I'm reverting these specific changes.


AlejandroRigau (approved):
I think I called this out in a previous PR. Thanks for catching that it accidentally went through

estohlmann (approved):

dustins:
> I think I called this out in a previous PR. Thanks for catching that it accidentally went through

Yep, you had called this out and I was trying to fix it quickly but for now it's better to just revert those changes.

douglas1850 (approved):

PR #157, opened 2024-06-19
Mention quota in access denied error

douglas1850:
*Description of changes:* When trying to start a training job with a GPU instance that you don't have a high enough service limit for, instead of giving an actually useful error boto3 will give a blanket `AccessDeniedException` and tell you that you lack CreateTrainingJob permissions.

This change will reword the user-friendly error message so that it also mentions the possibility of the customer hitting their service limits.

Also added a period to the displayed error.

### DIFF
--- a/backend/src/ml_space_lambda/utils/common_functions.py
+++ b/backend/src/ml_space_lambda/utils/common_functions.py
@@ -222,7 +222,7 @@ def generate_exception_response(e, status_code=400):
             },
             {
                 "Codes": ["AccessDeniedException"],
-                "FriendlyMessage": "An administrator or owner has restricted access to this resource or this account has hit its service limit for this resource. If you need access or to increase the limits, please contact a system administrator or owner of the resource for assistance",
+                "FriendlyMessage": "An administrator or owner has restricted access to this resource or this account has hit its service limit for this resource. If you need access or to increase the limits, please contact a system administrator or owner of the resource for assistance.",
             },
         ]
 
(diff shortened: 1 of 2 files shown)


backend/src/ml_space_lambda/utils/common_functions.py
```diff
@@ -222,15 +222,15 @@ def generate_exception_response(e, status_code=400):
...
+                "FriendlyMessage": "An administrator or owner has restricted access to this resource or this account has hit its service limit for this resource. If you need access or to increase the limits, please contact a system administrator or owner of the resource for assistance",
             },
         ]
 
         for errorType in ERROR_DICT:
             if e.response["Error"]["Code"] in errorType["Codes"] and (
                 "MatchStrings" not in errorType or any(error in error_msg for error in errorType["MatchStrings"])
             ):
-                e = f"{errorType['FriendlyMessage']} Full message: {error_msg}"
+                e = f"{errorType['FriendlyMessage']}. Full message: {error_msg}"
```
dustins, line 233: Are we sure all our error messages don’t have a period so this isn’t doubling up? I think it might be better to just change our error messages instead of inserting the period here.

douglas1850: good idea, will do

estohlmann: That or add some sort of conditional like

`errorType['FriendlyMessage'].endsWith(".") ? errorType['FriendlyMessage'] : errorType['FriendlyMessage'].`

dustins (approved):

estohlmann (approved):

PR #160, opened 2024-06-21
Updated workflows for new development target

dustins:
*Issue #, if available:*

*Description of changes:*
- reorganized workflows
- added new workflow for deploying to development environment


dustins:
I think I may be able to combine these two deployment workflows in the future but I am keeping them separate for the time being to test that the `main` branch doesn't deploy to our `development` environment even though that is listed as a trigger for this workflow. I want to make sure the branch protection feature of the github environment is working.

PR #163, opened 2024-06-21
Bugfix/hpo validation

douglas1850:
*Description of changes:* Form validation was broken for some of our HPO built-in algorithms.

It seems like the issue was with this conditional: 
```
if ((!hyperParameter.zValidator?.isOptional || value || (index === 1 && parameterRange.MinValue)) && !nanError ) {
```

The problem with it, is that `hyperParameter.zValidator` does not always have an `isOptional` property, and even when it does it's often nested below other properties.

Rather than adding more conditionals to check each possible nested property for `isOptional` I have updated the code so we just let `safeParse` do all of the hard work.

Things tested:

* required fields are behaving correctly
* if a min/max value is set, its sibling must also be set
* if a max is < the min or vice versa, it flags this
* handles letter values in number fields

I also updated the HPO Wizard so that when a job is submitted, the "Submit" button will show a loading icon.


frontend/src/entities/jobs/algorithms.tsx
```diff
@@ -67,6 +67,7 @@ export type Hyperparameter = {
     type: HyperparameterType;
     typeOptions: HyperparameterType[];
     scalingType?: string;
+    required: boolean;
```
estohlmann, line 70: Couldn't we just default this to false so we don't have to update all the fields below and just the ones we want to be required:

`required: boolean = false;`

douglas1850: Doesn't seem like I can do defaults like that: https://stackoverflow.com/questions/61862235/default-value-for-typescript-type-alias

I *could* set the default like this:
```
// Declare the type
export type SomeType = {
    typename: string;
    strength: number;
    radius: number;
    some_func: Function;
    some_other_stat: number;
}

// Create an object with all the necessary defaults
const defaultSomeType = {
    some_other_stat: 8
}

// Inject default values into your variable using spread operator.
const someTypeVariable: SomeType = {
  ...defaultSomeType,
  typename: 'name',
  strength: 5,
  radius: 2,
  some_func: () => {}
}
```

estohlmann: Interesting, I thought you could do this. No worries! I'll approve.

estohlmann (approved):

dustins (dismissed):

frontend/src/entities/jobs/hpo/create/training-definitions/training-job-definition.tsx
```diff
@@ -248,7 +247,7 @@ export function TrainingJobDefinition (props: TrainingJobDefinitionProps) {
                                         // - OR the field has a value
                                         // - OR this is the max field and the min field has a value
                                         // This evaluation avoids the issue where a blank value is evaluated as a 0 on a forced safeParse even for optional fields
-                                        if ((!hyperParameter.zValidator?.isOptional || value || (index === 1 && parameterRange.MinValue)) && !nanError ) {
+                                        if ((hyperParameter.required || value || (index === 1 && parameterRange.MinValue)) && !nanError ) {
```
dustins, line 250: I have a couple issues with what we're doing here. I don't like that you're breaking some validation out-of-band from Zod when the concept of optional exists within Zod. Shouldn't we be running the validator if it is present no matter what? If the value can be empty then our validator should be handling that.

I also don't think how it was working (by looking a property of the validator instead of just executing the validator) was the best way to go about it either for the reasons you've listed.

douglas1850: I think I see what you mean; we should just confirm that it has a `zValidator` property but then in the below lines where we actually invoke the validator we should let that confirm that it's valid. 

I'll try that and see if it works like I think it should

frontend/src/entities/jobs/hpo/create/training-definitions/training-job-definition.tsx
```diff
@@ -264,6 +258,13 @@ export function TrainingJobDefinition (props: TrainingJobDefinitionProps) {
                                                             ) +
                                                         ` (${index === 0 ? 'min value' : 'max value'})`,
                                                 });
+                                            } else if (value && ((index === 1 && !parameterRange.MinValue) || (index === 0 && !parameterRange.MaxValue))) {
```
dustins, line 261: Wouldn't this add an issue if MinValue/MaxValue is `0` even though that could be valid?

douglas1850: Screenshot 2024-06-24 at 4 43 04 PM

Like this?

dustins: Assuming `0`  is allowed in the range, if the `MaxRange = 1` (or whatever) is valid wouldn't it still show as invalid because `!parameterRange.MinValue` would evaluate to true?

douglas1850: To be clear these are all being seen as String values, so 0 is "0", if you're concerned about the number 0 being an un-truthy value

dustins (approved):

estohlmann (approved):

PR #169, opened 2024-06-25
Adding BE apis and CDK infrastructure to support CRUD on groups

estohlmann:
Adding BE APIs to manage Groups in MLSpace


backend/src/ml_space_lambda/enums.py
```diff
@@ -101,6 +103,7 @@ class Permission(Enum):
     PROJECT_OWNER = "MO"
     ADMIN = "PMO"
     ACTING_PMO = "actingPMO"
+    GROUP_OWNER = "GO"
```
douglas1850, line 106: I think we need to remove Group owner. For right now it seems like we want to lock down Group ownership to just admins (according to Cindy's notes), so I don't think this new role will give us anything.

douglas1850: Actually you know what, I feel like it'll be easier to just give an admin `GO` permissions when they create a group than to remove this and rip out all of the GroupOwner stuff, only to add it back in later if we want to support this for non-admins.

douglas1850 (approved):

backend/src/ml_space_lambda/enums.py
estohlmann: I was thinking we can add it right now and keep it similar to what we do for projects and then rip it out if we end up not wanting it before we call it feature complete.

README.md
```diff
@@ -134,56 +134,58 @@ cdk deploy --all
...
+| LAMBDA_RUNTIME                                 |                                                                                                                             The lambda runtime to use for MLSpace lambda functions and layers. This needs to be a python runtime available in the region in which MLSpace is being deployed.                                                                                                                              |                         Python 3.11 |
+| LAMBDA_ARCHITECTURE                            |                                                                           The architecture on which to deploy the MLSpace lambda functions. All lambda layers will also need to be built for the selected archiecture. You can do this by ensuring you run the `cdk deploy` command from a machine with the same architecture you're targeting.                                                                           |                                 x86 |
```
dustins, line 191: What is this change for? Best I can tell you added a couple parameters but the formatting of everything seems jacked now. Maybe it is just how github is showing it?

backend/src/ml_space_lambda/data_access_objects/group_user.py
```diff
@@ -0,0 +1,130 @@
...
+            group_name=dict_object["group"],
+            permissions=permissions,
+            role=dict_object.get("role", ""),
+        )
+
+
+class GroupUserDAO(DynamoDBObjectStore):
+    def __init__(self, table_name: Optional[str] = None, client=None):
+        self.env_vars = get_environment_variables()
+        table_name = table_name if table_name else self.env_vars["GROUP_USERS_TABLE"]
```
dustins, line 66: Looks like you added an entry to this to `EnvVariable` but didn't use it here.

backend/src/ml_space_lambda/data_access_objects/group_user.py
```diff
@@ -0,0 +1,130 @@
...
+
+class GroupUserModel:
+    def __init__(
+            self,
+            username: str,
+            group_name: str,
+            role: Optional[str] = None,
+            permissions: Optional[List[Permission]] = None,
+    ):
+        permissions = permissions if permissions else []
```
dustins, line 38: FWIW I think in these scenarios we can just do `permissions or []` and below `role or ""` to make it slightly more readable.

backend/src/ml_space_lambda/data_access_objects/group.py
```diff
@@ -0,0 +1,143 @@
...
+            dict_object["createdBy"],
+            dict_object.get("createdAt", None),
+            dict_object.get("lastUpdatedAt", None),
+        )
+
+
+class GroupDAO(DynamoDBObjectStore):
+    def __init__(self, table_name: Optional[str] = None, client=None):
+        self.env_vars = get_environment_variables()
+        table_name = table_name if table_name else self.env_vars["GROUPS_TABLE"]
```
dustins, line 68: This could also use the `EnvVariable` enum

dustins (approved):

README.md
estohlmann: I added the new tables `GROUPS_TABLE_NAME` and `GROUP_USERS_TABLE_NAME` to this, seems my IntelliJ reformatted the tables so that they are now the same length for every column.

PR #178, opened 2024-06-27
Bug: instance constraint issues

dustins:
*Issue #, if available:*

*Description of changes:*
The current notebook policy no longer grants permission for certain actions when `MANAGE_IAM_ROLES` is set to `true`. Instead, the notebook role inherits these permissions through the endpoint configuration and jobs instance constraint policies. As these policies are newly introduced, they were not included in the app-role when created with CDK. Consequently, only the notebook policy was added to the app-role, resulting in insufficient permissions.

- added instance constraint policies to CDK generated app role
- updated initial app config lambda to populate instance constraint policies
- updated tests

### DIFF
--- a/backend/src/ml_space_lambda/initial_app_config/lambda_function.py
+++ b/backend/src/ml_space_lambda/initial_app_config/lambda_function.py
@@ -18,7 +18,7 @@ import boto3
 
 from ml_space_lambda.app_configuration.policy_helper.notebook import update_instance_constraint_policies
 from ml_space_lambda.data_access_objects.app_configuration import AppConfigurationDAO
-from ml_space_lambda.enums import ServiceType
+from ml_space_lambda.enums import EnvVariable, ServiceType
 from ml_space_lambda.metadata.lambda_functions import get_compute_types
 from ml_space_lambda.utils.common_functions import generate_html_response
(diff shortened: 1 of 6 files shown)


backend/src/ml_space_lambda/initial_app_config/lambda_function.py
```diff
@@ -46,6 +47,7 @@ def lambda_handler(event, context):
     ]
 
     app_configuration_dao.update(config)
+    update_instance_constraint_policies(config, context)
```
douglas1850, line 50: Just checking, but what if DR isn't enabled, will this be able to add the required policies to the allowed instance types? 

If the answer is no, I suppose the customer just needs to make sure they have their IAM policies sorted or correctly right?

douglas1850 (approved):

backend/src/ml_space_lambda/initial_app_config/lambda_function.py
dustins: This won't be able to update if DR is enabled. I am going to update this with a condition to check that for good measure.

If they're not using DR it is up to them to have their policies in sync with the UI. Likely they'll just have permissive policies though.

PR #195, opened 2024-06-27
EMR detail page race condition fix

douglas1850:
*Description of changes:* When you create a cluster it will navigate you to the details page and immediately try to GET that cluster.

For reasons outside of MLSpace (nothing in our authorizer or lambda logs) this is returning a 403... probably because EMR is still setting up that cluster and it doesn't exist in the EMR ecosystem yet.

This fix will handle this issue by continuing to display a "loading" page while we refresh every 3 seconds and continuously try to fetch the cluster (in my testing it typically takes 3-9 seconds to populate).

### DIFF
--- a/frontend/src/entities/emr/detail/emr-clusters-detail.tsx
+++ b/frontend/src/entities/emr/detail/emr-clusters-detail.tsx
@@ -124,7 +124,7 @@ function EMRDetail () {
                 header={
                     <Header variant='h1'>
                         {' '}
-                        Cluster: {cluster.Name} ({clusterId}){' '}
+                        Cluster: {clusterName} ({clusterId}){' '}
                     </Header>
                 }
             >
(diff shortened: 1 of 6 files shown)


frontend/src/entities/emr/detail/emr-clusters-detail.tsx
```diff
@@ -67,13 +66,12 @@ function EMRDetail () {
         }
     }, [dispatch, basePath, projectName, clusterId, clusterName]);
 
-    // Refresh data in the background to keep state fresh
+    // Refresh data in the background to keep state fresh. While waiting for the EMR cluster to be ready, refresh every 3 seconds
     const isBackgroundRefreshing = useBackgroundRefresh(async () => {
         await dispatch(getEMRCluster(clusterId!));
-    }, [dispatch]);
+    }, [dispatch], true, clusterLoading ? 3 : window.env.BACKGROUND_REFRESH_INTERVAL);
```
dustins, line 72: Obviously this was on purpose but doesn't 3s seem a bit aggressive?

douglas1850: It typically took 3-9 seconds in testing for it to succeed

dustins (approved):

frontend/src/entities/emr/detail/emr-clusters-detail.tsx
dustins: I won't block on this.

estohlmann (approved):

PR #209, opened 2024-07-02
CRUD Groups without in app group user management

estohlmann:
Initial changes for CRUD on in app group management. 

User management in groups will come in a later PR, as will non-admin users being able to view what groups they are in.

demo

### DIFF
--- a/frontend/src/entities/group/create/group-create.tsx
+++ b/frontend/src/entities/group/create/group-create.tsx
@@ -14,9 +14,9 @@
  limitations under the License.
  */
 
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { z } from 'zod';
-import { issuesToErrors, useValidationReducer } from '../../../shared/validation';
+import { useValidationReducer } from '../../../shared/validation';
 import { useAppDispatch } from '../../../config/store';
 import NotificationService from '../../../shared/layout/notification/notification.service';
 import { useNavigate, useParams } from 'react-router-dom';
@@ -58,12 +58,12 @@ export function GroupCreate ({isEdit}: GroupCreateProperties) {
             .max(4000, {
                 message: 'Group description cannot be more than 4000 characters.',
             })
-            .regex(/[ -~]/, {
+            .regex(/^[ -~]+$/, {
                 message: 'Group description can only contain printable characters',
             }),
     });
 
-    const {state, setState, setFields, touchFields} = useValidationReducer(groupSchema, {
+    const {state, setState, errors, isValid, setFields, touchFields} = useValidationReducer(groupSchema, {
         validateAll: false,
         needsValidation: false,
         form: {
@@ -75,7 +75,7 @@ export function GroupCreate ({isEdit}: GroupCreateProperties) {
         formSubmitting: false,
     });
 
-    useMemo(() => {
+    useEffect(() => {
         if (isEdit && !initialLoaded) {
             setState({formSubmitting: true});
             dispatch(getGroup(groupName)).then((response) => {
@@ -94,17 +94,7 @@ export function GroupCreate ({isEdit}: GroupCreateProperties) {
             });
         }
         // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [isEdit, groupName, initialLoaded]);
-
-
-    state.formErrors = {};
-    const parseResult = groupSchema.safeParse(state.form);
-    if (!parseResult.success) {
-        state.formErrors = issuesToErrors(
-            parseResult.error.issues,
-            state.validateAll === true ? undefined : state.touched
-        );
-    }
+    }, [isEdit, groupName, initialLoaded, dispatch, navigate]);
 
     async function handleSubmit () {
         setState({formSubmitting: true});
@@ -169,7 +159,7 @@ export function GroupCreate ({isEdit}: GroupCreateProperties) {
                             onClick={() => {
                                 handleSubmit();
                             }}
-                            disabled={state.formSubmitting || !parseResult.success}
+                            disabled={state.formSubmitting || !isValid}
                         >
                             Submit
                         </Button>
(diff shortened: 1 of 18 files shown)


frontend/src/entities/group/create/group-create.tsx
```diff
@@ -0,0 +1,217 @@
...
+            })
+            .min(1, {
+                message: 'Group description cannot be empty',
+            })
+            .max(4000, {
+                message: 'Group description cannot be more than 4000 characters.',
+            })
+            .regex(/[ -~]/, {
+                message: 'Group description can only contain printable characters',
+            }),
```
dustins, line 63: IMHO I think this convention is very confusing to read. It looks like ` `, `-`, or `~`. But barring that I think this should probably be `/^[ -~]$/` so every character is not a printable character.

frontend/src/entities/group/create/group-create.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+
+    state.formErrors = {};
+    const parseResult = groupSchema.safeParse(state.form);
+    if (!parseResult.success) {
+        state.formErrors = issuesToErrors(
+            parseResult.error.issues,
+            state.validateAll === true ? undefined : state.touched
+        );
+    }
```
dustins, line 107: This block is probably unnecessary. The `useValidationReducer` function above also runs `isValid` and `errors` and is essentially doing [the same thing](https://github.com/awslabs/mlspace/blob/main/frontend/src/shared/validation/index.ts#L296-L303) inside the method

frontend/src/entities/group/create/group-create.tsx
```diff
@@ -0,0 +1,217 @@
...
+    async function handleSubmit () {
+        setState({formSubmitting: true});
+        let response = null;
+        try {
+            if (!isEdit) {
+                response = await dispatch(createGroup(state.form));
+            } else {
+                response = await dispatch(updateGroup(state.form));
+            }
+            if (response.payload?.status === 200) {
```
dustins, line 118: So we don't have to do a bunch of explicit status checks everywhere we've been trying to standardize on [isFulfilled](https://redux-toolkit.js.org/api/matching-utilities#isfulfilled). You can see an [example here](https://github.com/awslabs/mlspace/blob/65547456b15f42268766527eba88cf0e26365ab1/frontend/src/entities/notebook/notebook.actions.tsx#L169-L174)

frontend/src/entities/group/detail/group-detail.actions.tsx
```diff
@@ -0,0 +1,129 @@
...
+                            setModalState({
+                                ...modalState,
+                                visible: false,
+                            });
+                            notificationService.showActionNotification(
+                                'delete group',
+                                `Group ${groupName} deleted.`,
+                                result
+                            );
+                            if (result.type.endsWith('/fulfilled')) {
```
dustins, line 118: Same thing here with `isFullfilled`

frontend/src/entities/group/create/group-create.tsx
```diff
@@ -0,0 +1,217 @@
...
+                            description: response.payload?.data.group?.description || '',
+                        },
+                        formSubmitting: false
+                    });
+                } else {
+                    navigate('/404');
+                }
+            });
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
dustins, line 96: Do these exhaustive checks have to be disabled everywhere or is that just from copying and pasting?

estohlmann: this particular one was needed due to setState causing this to continuously reload if it was in the dependency list. I was able to remove the others in `group-detail.tsx` though.

frontend/src/entities/group/create/group-create.tsx
estohlmann: Sure, no preference from me as long as the message is clear. I updated. 

I had copied this from the project description in the `project-create.tsx` file.

frontend/src/entities/group/create/group-create.tsx
estohlmann: So it seems isFulfilled just validates that the thunkRequest finished and we will need to keep this logic or add logic in the reducer to check for response statuses and reject appropriately. I tested this and then navigated to edit a non-existent group, the request returned a 404 but isFulfilled still returned `true`.

frontend/src/entities/group/create/group-create.tsx
estohlmann: Thanks, I hadn't used this before. Still learning the libraries.

dustins (approved):

dustins (approved):

frontend/src/entities/group/detail/group-detail.tsx
```diff
@@ -0,0 +1,103 @@
...
+                    columns={1}
+                    header='Group details'
+                    actions={GroupDetailActions()}
+                    info={groupDetails}
+                    loading={!initialLoaded}
+                />
+                {/*<Container>*/}
+                <Table
+                    tableName='Group member'
+                    // tableType={tableType}
```
douglas1850, line 87: Should these just be removed? Same with the commented out Container on line 84

douglas1850: And line 97

frontend/src/entities/group/group.reducer.ts
```diff
@@ -0,0 +1,82 @@
...
+    const requestUrl = `/group/${groupName}/users`;
+    return axios.get<IGroupUser[]>(requestUrl);
+});
+
+export const GroupSlice = createSlice({
+    name: 'group',
+    initialState,
+    extraReducers (builder) {
+        builder
+            .addMatcher(isFulfilled(getAllGroups), (state, action) => {
```
douglas1850, line 60: Do we need a matcher for `getGroup` or `getGroupUsers`?

frontend/src/entities/group/detail/group-detail.tsx
estohlmann: I left them because they will be needed when I add user management in the next task. Since this isn't going into develop I figured it would be okay.

douglas1850: ah okay

douglas1850 (approved):

PR #210, opened 2024-07-03
Add group permissions to a user when added to a group

douglas1850:
*Description of changes:* When a user is added to a group, update their user policy so it has permissions to see datasets related to that Group. Inversely if a user is removed from a group, or a group is deleted, then remove the group permissions from a user.


backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -322,6 +212,43 @@ def add_iam_role(self, project_name: str, username: str) -> str:
...
 
+    # Create or update the user policy
+    def update_user_policy(
+        self,
+        username: str,
+    ) -> str:
+        user_policy_name = f"{IAM_RESOURCE_PREFIX}-user-{username}"
+        aws_account = self.sts_client.get_caller_identity()["Account"]
+        user_policy_arn = f"arn:{self.aws_partition}:iam::{aws_account}:policy/{user_policy_name}"
+        self._check_name_length(IAMResourceType.POLICY, user_policy_name)
```
dustins, line 223: I know you didn't write this but…what is the point of this? It throws an exception if the policy name is too long without any way for the app/user to do anything about it because AFAIK we don't have a way to change our username. I think we need to really brainstorm how we can move away from this naming scheme.

dustins (approved):

estohlmann (approved):

PR #211, opened 2024-07-04
FE in app group user management

estohlmann:
Added the ability for admins to add and remove users to/from groups in MLSpace.

demo


estohlmann:
Added the ability for admins to create groups and pre-select users to be added to the group.

CreateGroupWithUsers

douglas1850 (approved):

frontend/src/entities/group/detail/group-detail-user.actions.tsx
```diff
@@ -0,0 +1,121 @@
...
+            disabled={!selectedUser}
+            onItemClick={(e) => GroupDetailUserActionHandler(e, dispatch, modalState as ModalProps, setModalState, selectedUser)}
+        >
+            Actions
+        </ButtonDropdown>
+    );
+}
+
+const GroupDetailUserActionHandler = async (
+    e: any,
```
dustins, line 85: I know this is done other places in the code but `any` should be `CustomEvent<ButtonDropdownProps.ItemClickDetails>`

frontend/src/entities/group/detail/group-detail-user.actions.tsx
```diff
@@ -0,0 +1,121 @@
...
+            >
+                Add User
+            </Button>
+        </SpaceBetween>
+    );
+}
+
+function GroupDetailUserActionsButton (navigate: NavigateFunction, dispatch: Dispatch, props?: any) {
+    const selectedUser: IGroupUser = props?.selectedItems[0];
+    const items = [];
```
dustins, line 58: This could be `const items: ButtonDropdownProps.Item[] = []`, I'm surprised our linter didn't block on this.

frontend/src/entities/group/detail/group-detail-user.actions.tsx
```diff
@@ -0,0 +1,121 @@
...
+) => {
+    const notificationService = NotificationService(dispatch);
+
+    switch (e.detail.id) {
+        case 'removeFromGroup':
+            dispatch(
+                setDeleteModal({
+                    resourceName: 'Group User',
+                    resourceType: 'groupUser',
+                    postConfirm: () => dispatch(getGroupUsers(selectedUser.group)),
```
dustins, line 99: Again our linter is not doing the work it needs to here I guess. `group` could is optional on `IGroupUser` so this is not a safe call. If there is no scenario where `group` is going to actually be undefined then lets adjust the type.

frontend/src/entities/group/detail/group-detail.tsx
```diff
@@ -26,17 +26,19 @@ import GroupDetailActions from './group-detail.actions';
...
+import { GroupDetailUserActions } from './group-detail-user.actions';
+import { DocTitle } from '../../../shared/doc';
 
 export function GroupDetail () {
     const dispatch = useAppDispatch();
     const navigate = useNavigate();
     const {groupName} = useParams();
     const [group, setGroup] = useState<IGroup>(null);
-    const [groupUsers, setGroupUsers] = useState<IGroupUser>([]);
+    const groupUsers: IGroupUser[] = useAppSelector((state) => state.group.currentGroupUsers) || [];
```
dustins, line 37: FWIW the project _tries_ to [write selectors](https://github.com/awslabs/mlspace/blob/develop/frontend/src/entities/configuration/configuration-reducer.ts#L150-L154) so we're not doing so much of this filtering inline and then we can define the return type so we're not having to type it ourselves. Although the example I linked to probably isn't following the best practice of prefixing it with `select...`

frontend/src/entities/group/detail/group-detail.tsx
```diff
@@ -47,6 +49,7 @@ export function GroupDetail () {
             dispatch(getGroup(groupName)).then((response) => {
```
dustins, line 49: This isn't specific to this PR but looks like `getGroup` has the wrong return type defined. It is declared as returning `IGroup` but seems like it returns `{group: IGroup, permissions: Permission[]}`

frontend/src/entities/group/detail/group-detail-user.actions.tsx
```diff
@@ -0,0 +1,121 @@
...
+                            setModalState({
+                                ...modalState,
+                                visible: false,
+                            });
+                            notificationService.showActionNotification(
+                                'remove group user',
+                                `User ${selectedUser.user} removed from ${selectedUser.group}.`,
+                                result
+                            );
+                        }),
```
dustins, line 111: The typing of this is so wacky I don't even have an immediate suggestion on how to fix it. Part of it is just how `onConfirm` was typed to begin with.

frontend/src/entities/group/user/group-user-functions.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export async function addUsersToGroup (
+    dispatch: ThunkDispatch<any, any, Action>,
+    groupName: string,
+    users: IUser[]
+) {
+    const notificationService = NotificationService(dispatch);
+    const request: AddGroupUserRequest = {
+        groupName: groupName,
+        usernames: users.map((user) => user.username)
```
dustins, line 29: This is invalid since `user.username` is optional. We should correct the IUser model or add `.filter(Boolean)` so we filter out undefined values.

dustins (approved):

dustins (approved):

PR #214, opened 2024-07-05
Remove default user and group owner

estohlmann:
Removed Default user upon group creation so that admins can create empty groups if they choose.

I also removed the concept of `GroupOwners` since this won't be used upon feature rollout. We can add this back if we end up thinking this is needed at a later time. All admins will be able to act in this capacity.


estohlmann:
We can address these checks at the end of the feature, the checks ran because I had pointed this at develop to start.

douglas1850 (approved):

backend/src/ml_space_lambda/group/lambda_functions.py
```diff
@@ -53,17 +53,6 @@ def _add_group_user(group_name: str, username: str, permissions: Optional[List[P
         raise e
 
 
-def _total_group_owners(group_name: str):
```
douglas1850, line 56: Just to make sure I'm understanding correctly - all Admins are basically owners for all groups, right? Just making sure we can never have a situation where an Admin removes themself from a Group as a member and now we have empty groups floating around that nobody can see in the UI.

estohlmann: Correct, admins can view all groups all the time

PR #218, opened 2024-07-09
Group dataset creation

douglas1850:
*Description of changes:* Users can now create datasets that are associated directly with Groups and will be viewable by anyone else in that Group.

### What's done
* dataset creation - the new dataset is created and added to the dataset table, and uploaded files are stored correctly in S3
* edit/delete/get - all tested and working
* select existing datasets by using the Dataset Selector
* create new Group datasets inline


backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -324,6 +326,15 @@ def lambda_handler(event, context):
                                 policy_statement["Effect"] = "Allow"
                     elif target_type == DatasetType.PRIVATE and username == target_scope:
                         policy_statement["Effect"] = "Allow"
+                    elif target_type == DatasetType.GROUP:
+                        if Permission.ADMIN in user.permissions:
+                            policy_statement["Effect"] = "Allow"
```
dustins, line 331: Just seeking some input…

We're doing a check like this…or at least `Permission.ADMIN in user.permissions` 10+ times in this file. Is there a scenario where we don't want the Admin to have access to something? We could clean up this file a bit by just putting a check for that at the beginning and not doing this check everywhere.

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -324,6 +326,15 @@ def lambda_handler(event, context):
...
+                        else:
+                            # TODO: check if this user is a member of the Group
```
dustins, line 333: Is this still a TODO? Looks like you did the damn thing.

backend/test/dataset/test_create_dataset.py
```diff
@@ -88,7 +89,7 @@ def test_create_dataset_success(mock_s3, mock_dataset_dao, dataset_type: str, sc
 
     dataset_location = f's3://{TEST_ENV_CONFIG["DATA_BUCKET"]}/{directory_name}'
 
-    test_dataset = generate_dataset_model(event=mock_event, scope=scope, dataset_location=dataset_location)
+    test_dataset = generate_dataset_model(event=mock_event, scope=scope, dataset_location=dataset_location, type=dataset_type)
```
dustins, line 92: Not that we expect it to change anything but for good measure this is a parameterized test so we should add GROUP to it probably right?

frontend/src/entities/dataset/dataset.service.tsx
```diff
@@ -139,12 +144,11 @@ export async function uploadResources (datasetContext: DatasetContext, resourceO
     const failedUploads: string[] = [];
     let stopUpload = false;
 
+    console.log(`Uploading ${resourceObjects.length} resources`);
```
dustins, line 147: We should remove this

frontend/src/entities/dataset/detail/dataset-detail.tsx
```diff
@@ -68,15 +68,15 @@ function DatasetDetail () {
             setBreadcrumbs([
                 getBase(projectName),
                 { text: 'Datasets', href: `#${basePath}/dataset` },
-                { text: `${name}`, href: `#${basePath}/dataset/${scope}/${name}` },
+                { text: `${name}`, href: `#${basePath}/dataset/${type}/${scope}/${name}` },
             ])
         );
-        dispatch(getDatasetByScopeAndName({ scope: scope, name: name }))
+        dispatch(getDatasetByScopeAndName({ type: type, scope: scope, name: name }))
```
dustins, line 74: `getDatasetsByScopeAndName` probably isn't a good name anymore :x

frontend/src/modules/dataset/dataset-browser.reducer.tsx
```diff
@@ -147,16 +147,7 @@ export type DatasetBrowserState = {
...
-        const { datasetContext, projectName, username } = props;
-        let scope: string | undefined = datasetContext.type;
-
-        switch (datasetContext.type) {
-            case DatasetType.PROJECT:
-                scope = projectName;
-                break;
-            case DatasetType.PRIVATE:
-                scope = username;
-        }
```
dustins, line 159: Are you super-extra-100%-definitely sure you can remove this? Like I think [here](https://github.com/awslabs/mlspace/blob/develop/frontend/src/modules/dataset/dataset-selector.tsx#L52-L55) it is called with the results from [datasetFromS3Uri](https://github.com/awslabs/mlspace/blob/develop/frontend/src/shared/util/dataset-utils.ts#L21) in which scope might be empty for the global scope. So I think unless we modify `datasetFromS3Uri` (which might be possible) we can't assume `.scope` is always defined.

frontend/src/shared/util/dataset-utils.ts
```diff
@@ -16,7 +16,7 @@
 
 import { DatasetType, IDataset } from '../model';
 
-export type DatasetContext = IDataset & Required<Pick<IDataset, 'name' | 'type'>>;
+export type DatasetContext = IDataset & Required<Pick<IDataset, 'name' | 'type' | 'scope'>>;
```
dustins, line 19: See my comment above about `scope` for this function. A pox on your linter's family for not catching this but this is invalid I'm pretty sure. It's fine to say the result will have a field `scope` but it needs to be a `string` and this function is potentially setting `undefined` by saying `matches?.scope`.

frontend/src/modules/dataset/dataset-browser.utils.tsx
```diff
@@ -161,7 +161,7 @@ function tablePropertiesForScopes (state: DatasetBrowserState, updateDatasetCont
             cell (item) {
                 return (
                     <Link data-cy={item.name} onFollow={() => {
-                        updateDatasetContext({ type: item.type }, '', false);
+                        updateDatasetContext({ ...state.datasetContext, type: item.type }, '', false);
```
dustins, line 164: If nothing else in this file was changed why is this needed?

frontend/src/entities/dataset/create/dataset-create.tsx
```diff
@@ -107,6 +111,32 @@ export function DatasetCreate () {
...
     }, [dispatch, basePath, projectName]);
 
+    useEffect(() => {
+        dispatch(getAllGroups());
+    }, [dispatch]);
+
+    function generateOptions () {
+        // Standard options always available
+        const options: { label: string; value?: string; options?: any[] }[] = [
+            { label: 'Global', value: 'global' },
```
dustins, line 121: Maybe lets make these `{label: initCap(DatasetType.GLOBAL), value: DatasetType.GLOBAL}`

frontend/src/entities/dataset/create/dataset-create.tsx
```diff
@@ -107,6 +111,32 @@ export function DatasetCreate () {
...
+            { label: 'Private', value: 'private' },
+        ];
+
+        if (projectName) {
+            options.push({ label: 'Project', value: 'project' });
```
dustins, line 126: Probably should do the same as above.

frontend/src/entities/dataset/create/dataset-create-functions.tsx
```diff
@@ -14,15 +14,18 @@
...
 */
 
-import { IDataset } from '../../../shared/model/dataset.model';
+import { DatasetType, IDataset } from '../../../shared/model/dataset.model';
 import { determineScope } from '../dataset.service';
 
 export const createDatasetFromForm = (form: any, projectName: string, username: string): IDataset & Required<Pick<IDataset, 'name' | 'description' | 'type' | 'format' | 'scope'>> => {
+    if (form.type.startsWith(DatasetType.GROUP)) {
+        form.type = DatasetType.GROUP;
+    }
```
dustins, line 23: <jackie-chan-why /> ?

frontend/src/entities/dataset/dataset.service.tsx
```diff
@@ -121,13 +123,16 @@ export const fetchPresignedURL = async (s3Key: string) => {
...
+    groupName: string | undefined,
     username: string
 ): string => {
     switch (type) {
         case DatasetType.GLOBAL:
             return DatasetType.GLOBAL;
         case DatasetType.PROJECT:
             return projectName!;
+        case DatasetType.GROUP:
+            return groupName!;
```
dustins, line 135: This `!` isn't go unnoticed, I just don't have a nicer way to clean that up right now O_o

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: I *did* do the damn thing didn't I?

frontend/src/entities/dataset/detail/dataset-detail.tsx
douglas1850: `getDatasetByScope,Name,andFriends` is my vote

douglas1850: `getDataset` seems available, so I'll go with that for brevity.

frontend/src/modules/dataset/dataset-browser.utils.tsx
douglas1850: Hmm I was trying to get the browsers/selectors working, I think I just missed them in reverting my changes

frontend/src/entities/dataset/create/dataset-create-functions.tsx
douglas1850: I don't even remember writing this... I don't see the point of it either, will remove it

frontend/src/modules/dataset/dataset-browser.reducer.tsx
douglas1850: Hmm I think this was also from me tweaking the browsers/selectors, will revert this

frontend/src/modules/dataset/dataset-browser.reducer.tsx
```diff
@@ -171,7 +162,7 @@ export const getDatasetContents = createAsyncThunk(
             searchParams.set('prefix', datasetContext.location);
         }
 
-        return axios.get<ListFilesResponse>(`/dataset/${scope}/${datasetContext.name}/files?${searchParams.toString()}`).then((response) => {
+        return axios.get<ListFilesResponse>(`/dataset/${datasetContext.type}/${datasetContext.scope}/${datasetContext.name}/files?${searchParams.toString()}`).then((response) => {
```
douglas1850, line 165: Going to revert this too - it's going to be broken for now since the API is expecting scope and type in the param, but I'll address this is in the follow-up PR

backend/test/dataset/test_create_dataset.py
douglas1850: Yeah good idea

backend/src/ml_space_lambda/authorizer/lambda_function.py
douglas1850: What did you have in mind? Finding all of the places we do this check then condensing it into one conditional further up in the file? Like basically early on in this auth logic we'd first check if the user is an admin, then check if the requested resource is in some list of request resources that admins always have access to?

One argument in favor of the status quo is that it's very easy to follow. The logic is split up by request method and path params and I can easily find all of the dataset logic in one area, all of the job logic in one area, etc.

backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -187,11 +189,13 @@ def create_dataset(event, context):
     dataset_name = body.get("datasetName")
     env_variables = get_environment_variables()
 
+    print(f"Event: {event}")
```
douglas1850, line 192: Missed this, will remove on next commit

frontend/src/modules/dataset/dataset-inline-create.tsx
```diff
@@ -65,9 +65,10 @@ export function DatasetInlineCreate (props: DatasetInlineCreateProps) {
                 case DatasetType.PROJECT:
                     scope += `/${projectName}`;
                     break;
+                //TODO: need a GROUP case - requires a group name
```
douglas1850, line 68: Addressing this in follow-up PR for browser/selector changes

frontend/src/modules/dataset/dataset-selector.tsx
```diff
@@ -66,6 +66,7 @@ export function DatasetResourceSelector (props: DatasetResourceSelectorProps) {
                     case DatasetType.PROJECT:
                         scope = `${projectName}`;
                         break;
+                    //TODO: need groupName for a Group scope
```
douglas1850, line 69: Addressing this in follow-up PR for browser/selector changes

backend/src/ml_space_lambda/authorizer/lambda_function.py
dustins: What don't admins have access too? I feel like at the top of the file we can just say if you're an admin, allow it. Just being allowed to access the things doesn't change what we show admins in the UI. But I think they _should_ have access all resources right?

douglas1850: I thought private datasets, but if I'm wrong and admins can touch literally everything then that sounds like a good idea

dustins (approved):

frontend/src/modules/dataset/dataset-inline-create.tsx
```diff
@@ -29,8 +31,9 @@ export type DatasetInlineCreateProps = {
 
 export function DatasetInlineCreate (props: DatasetInlineCreateProps) {
     const {username, projectName, onChange} = props;
+    const groups: IGroup[] = useAppSelector((state) => state.group.allGroups);
 
-    const createOptions = enumToOptions(DatasetType, true).filter((option) => option.value !== DatasetType.GLOBAL);
+    //const createOptions = enumToOptions(DatasetType, true).filter((option) => option.value !== DatasetType.GLOBAL);
```
estohlmann, line 36: remove?

estohlmann (approved):

frontend/src/modules/dataset/dataset-inline-create.tsx
douglas1850: remove.

douglas1850, line 36:
```suggestion

```

PR #221, opened 2024-07-11
Group Cleanup

estohlmann:
*Description of changes:*

- Cleaning up groups so that admins pull back personal groups under personal
- Updating so that edit/update is consistent
- Pulling back all users navigating to create group page

### DIFF
--- a/backend/src/ml_space_lambda/group/lambda_functions.py
+++ b/backend/src/ml_space_lambda/group/lambda_functions.py
@@ -136,7 +136,7 @@ def remove_user(event, context):
 @api_wrapper
 def list_all(event, context):
     user = UserModel.from_dict(json.loads(event["requestContext"]["authorizer"]["user"]))
-    if is_admin_get_all(user, event.get("queryStringParameters", {})):
+    if is_admin_get_all(user, event):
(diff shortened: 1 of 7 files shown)


douglas1850 (approved):

backend/src/ml_space_lambda/group/lambda_functions.py
```diff
@@ -135,7 +136,7 @@ def remove_user(event, context):
 @api_wrapper
 def list_all(event, context):
     user = UserModel.from_dict(json.loads(event["requestContext"]["authorizer"]["user"]))
-    if Permission.ADMIN in user.permissions:
+    if is_admin_get_all(user, event.get("queryStringParameters", {})):
```
dustins, line 139: Because this is such a specific function maybe it makes sense to do the whole `event.get("queryStringParameters", {})` inside this method so we can just pass `event`. That will help clean up our code more and it already seems unlikely we'd be suing this function in other contexts. But then we'd have this beautiful `if is_admin_get_all(user, event):` instead and we can deal with the complexity in the function.

dustins (approved):

PR #245, opened 2024-07-23
Multi group datasets

douglas1850:
*Description of changes:* Group datasets can now be shared between multiple groups.

* when creating a dataset, a user can select multiple groups in a multiselect component that only appears when the dataset type is GROUP
* if a user is a member of multiple groups that have the dataset shared with them, de-duping has been handled (both in terms of the list UI as well as IAM resources for the user policy)
* when a group is deleted its group-dataset entries are deleted as well
* the detail page for a group now displays which groups a dataset has been shared with


backend/src/ml_space_lambda/data_access_objects/dataset.py
```diff
@@ -38,6 +38,7 @@ def __init__(
         created_by: str,
         created_at: Optional[float] = None,
         last_updated_at: Optional[float] = None,
+        groups: Optional[List] = [],
```
dustins, line 41: Do we want this as `Optional` if we're always defaulting it to `[]`? In what scenario do we expect it to be `None` instead of at worst an empty list?

backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -124,6 +124,10 @@ def get(event, context):
     dataset_name = event["pathParameters"]["datasetName"]
 
     dataset = dataset_dao.get(scope, dataset_name)
+    if scope == DatasetType.GROUP:
+        groups = group_dataset_dao.get_groups_for_dataset(dataset_name)
+        for group in groups:
+            dataset.groups.append(group.group)
```
dustins, line 130: Something to be cognizant of is that we're returning this dataset with the `groups` field filled in. Although it doesn't currently happen (to my knowledge) the `edit` method above takes the dataset object and persists it as-is. If this ever came back with these groups the next time it is fetched we'd be appending duplicates groups. At a minimum I think we should be doing something like `dataset.groups = list(set([group.name for group in groups]))`

frontend/src/entities/dataset/create/dataset-create.tsx
```diff
@@ -282,17 +289,40 @@ export function DatasetCreate () {
                                 <Select
                                     data-cy='dataset-type-select'
                                     selectedOption={{
-                                        label: state.form.groupName ? state.form.groupName : initCap(state.form.type || ''),
+                                        label: initCap(state.form.type || ''),
```
dustins, line 292: Pretty sure `state.form.type` can never be undefined so we don't need the `|| ''`

frontend/src/entities/dataset/create/dataset-create.tsx
```diff
@@ -127,17 +130,21 @@ export function DatasetCreate () {
...
-            options.push({ label: 'Groups', options: groupLabels});
+        if (groups) {
+            options.push({ label: initCap(DatasetType.GROUP), value: DatasetType.GROUP });
         }
 
         return options;
     }
 
+    function generateGroupOptions () {
+        const groupOptions: { label: string; value: string }[] = [];
```
dustins, line 141: Can we just use `SelectProps.Options[]` instead?

dustins:
We talked about being able to add/remove groups from a dataset but I don't see it here. Is that coming in a later point release?

douglas1850:
My latest commit allows us to add/remove groups from a group dataset, so we should be good on that front now

backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -107,6 +107,31 @@ def edit(event, context):
...
+    if dataset.type == DatasetType.GROUP:
+        # get the new list of groups that have this dataset shared with them.
+        # this list may be adding or removing existing groups from this dataset
+        new_groups_list = body.get("groups", [])
+        # get the current list of groups with access to this dataset so we can determine which groups were added/removed
+        groups = group_dataset_dao.get_groups_for_dataset(dataset_name)
+        old_group_names = []
+        group_difference = []
+
+        # Check for existing groups that are no longer in the group list
```
estohlmann, line 119: I think we cold be more efficient here by looping through by merging the new_groups_list and the groups list without dups and then looping through that once vs having two different loops. 

We can then do an if/else if to add/remove/ignore

backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -259,6 +294,12 @@ def create_dataset(event, context):
         if dataset_created:
             dataset_dao.delete(dataset_scope, dataset_name)
 
+        if group_dataset_created:
+            for group in groups:
+                group_dataset_dao.delete(dataset_name, group)
+                group_dataset_created = True
```
estohlmann, line 300: Remove

backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -259,6 +294,12 @@ def create_dataset(event, context):
         if dataset_created:
             dataset_dao.delete(dataset_scope, dataset_name)
```
estohlmann, line 295: Would we want to clean up group_datasets before deleting the dataset? This isn't relational so it doesn't matter but might make more sense

backend/src/ml_space_lambda/data_access_objects/dataset.py
douglas1850: Fair point, I'll just change it to defaulting to an empty list and remove the Optional bit

backend/src/ml_space_lambda/dataset/lambda_functions.py
douglas1850: So this actually isn't a problem because of how we update the DDB entry in the `edit` function:

```
exp_values = json.loads(
            dynamodb_json.dumps(
                {
                    ":description": dataset.description,
                    ":lastUpdatedAt": time.time(),
                    ":name": name,
                    ":scope": scope,
                }
            )
        )
```

We never load the `groups` field into the update so it never makes it to the table.

backend/src/ml_space_lambda/dataset/lambda_functions.py
douglas1850: I don't think it matters either way 🤷🏻‍♂️ I was just trying to maintain the order of

1. we create the dataset
2. we create the group_dataset rows

if things go wrong
1. we delete the dataset
2. we delete the group_dataset rows

frontend/src/entities/dataset/create/dataset-create.tsx
douglas1850: Seems crazy enough to work

backend/src/ml_space_lambda/dataset/lambda_functions.py
douglas1850: I was thinking through that earlier too... I could loop through both lists to create a list/set of the shared values, then loop again to determine what was added or removed... but this was easier and it seemed like it'd be the same time complexity overall.

estohlmann (dismissed):

backend/src/ml_space_lambda/data_access_objects/dataset.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         created_by: str,
         created_at: Optional[float] = None,
         last_updated_at: Optional[float] = None,
-        groups: Optional[List] = [],
+        groups: list = [],
```
dustins, line 41:
```suggestion
        groups: List[str] = [],
```

estohlmann (approved):

dustins (approved):

PR #246, opened 2024-07-26
Added ability to associate Groups with Projects

dustins:
*Issue #, if available:*

*Description of changes:*
- Added ability for Groups to be associated with Projects
- Groups can be given "owner" permissions of a group
- Users can still be directly associated with a Project and will take on the highest permissions available


backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -166,7 +164,7 @@ def lambda_handler(event, context):
                     and "username" in path_params
                     and request_method in ["GET", "PUT", "DELETE"]
                 ):
-                    if Permission.ADMIN in user.permissions:
+                    if IS_ADMIN:
```
douglas1850, line 167: Per our discussion in standup it sounds like we'll just move the IS_ADMIN check to the beginning of the authorizer

backend/src/ml_space_lambda/group/lambda_functions.py
```diff
@@ -121,11 +125,31 @@ def group_datasets(event, context):
...
+
+                if env_variables[EnvVariable.MANAGE_IAM_ROLES]:
+                    for project_group in project_groups:
+                        iam_role_arn = iam_manager.get_iam_role_arn(project_group.project, username)
+                        if iam_role_arn is None:
+                            iam_role_arn = iam_manager.add_iam_role(project_group.project, username)
+
+                iam_manager.update_user_policy(username)
+            else:
+                raise Exception("Username specified is not associated with an active user.")
```
douglas1850, line 152: does raising an exception here halt execution for the other users? 

This might be fine if we want this operation to be atomic, but it could also be annoying if the user just wants to add valid users and be alerted that at least one user failed to be added. I think we should at least raise the exception with the offending user's username so the error displayed in MLS lets the admin know which user is causing problems.

backend/src/ml_space_lambda/data_access_objects/project_group.py
```diff
@@ -0,0 +1,125 @@
...
+            "permissions": serialize_permissions(self.permissions),
+        }
+
+    @staticmethod
+    def from_dict(dict_object: dict) -> ProjectGroupModel:
+        permissions = [Permission(entry) for entry in dict_object.get("permissions", [])]
+        return ProjectGroupModel(
+            group_name=dict_object["group"],
+            project_name=dict_object["project"],
+            permissions=permissions,
```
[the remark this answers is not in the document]
dustins: Definitely a different PR but I think we should keep this as an array for the time being to match the rest of the project. Although we're reducing the number of permissions we have we don't know if we want to fully get rid of this concept of multiple permissions which would limit ourselves from expanding the types in the future.

backend/src/ml_space_lambda/authorizer/lambda_function.py
dustins: > I think we should hold off on bigger authorizer changes of short circuiting if they’re an admin. Some places in the authorizer set a variable on response_context (primarily the project name) and short circuiting evaluation would prevent that. I don’t immediately have a great solution for that either…

backend/src/ml_space_lambda/group/lambda_functions.py
dustins: You're right it would. I just folded back in [this code](https://github.com/awslabs/mlspace/pull/246/files/55a3316b3d5d6fa3e4149c5a242ed99221433afe#diff-cc7e87a7a0f5f44978d1b8428b06133c092432c8cd6b435298518025bc2e2042L46-L60) so we had this same behavior before but it isn't a great customer experience. I think the correct course of action is to validate all the users exist first and only then do the updates. This way we're at least trying to avoid putting things into a partial state and we can inform the requester what users are problematic.

douglas1850 (approved):

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -176,31 +174,34 @@ def lambda_handler(event, context):
...
                             )
-                            or (request_method in ["PUT", "DELETE"] and len(path_params) == 2 and "username" in path_params)
-                        ) and (
-                            (project_user and Permission.PROJECT_OWNER not in project_user.permissions)
-                            and Permission.ADMIN not in user.permissions
-                        ):
+                            or (
+                                request_method in ["PUT", "DELETE"]
+                                and len(path_params) == 2
+                                and ("username" in path_params or "groupName" in path_params)
```
estohlmann, line 194: Yeah, this is disgusting. 

Would really like to see about adding this in the individual `lambda_functions.py` files or maybe we break this out so that we are handling each entity in it's own method. Having groups combined with users is just going to lead some bugs.

Can be a future enhancement but let's get something in the backlog.

backend/src/ml_space_lambda/data_access_objects/project_group.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(
+        self,
+        group_name: str,
+        project_name: str,
+        permissions: Optional[List[Permission]] = None,
+    ):
+        permissions = permissions if permissions else []
+        self.group_name = group_name
+        self.project = project_name
+        self.permissions = permissions
```
estohlmann, line 41: can't we combine line 38 and this for something like:
`self.permissions = permissions if permissions else []`

backend/src/ml_space_lambda/data_access_objects/project_group.py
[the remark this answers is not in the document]
estohlmann: agreed, this is more future looking. I could see us eventually having more than admin permissions depending on customer asks.

backend/src/ml_space_lambda/data_access_objects/project_group.py
dustins: Better yet I think I'll just default the parameter to `[]` so if `None` is passed in we get the desired effect.

backend/src/ml_space_lambda/group/lambda_functions.py
```diff
@@ -191,11 +228,24 @@ def delete(event, context):
     # Delete users associations and group itself
     to_delete_group_users = group_user_dao.get_users_for_group(group_name)
 
+    # Remove group from all projects
+    for project_group in project_group_dao.get_projects_for_group(group_name):
+        project_group_dao.delete(project_group.project, group_name)
+
+        if env_variables[EnvVariable.MANAGE_IAM_ROLES]:
+            for group_user in to_delete_group_users:
+                if not is_member_of_project(group_user.user, project_group.project):
```
estohlmann, line 237: I think we want to remove the `not` here so that we are updating the policies and removing the project role from users that are in the project

backend/src/ml_space_lambda/user/lambda_functions.py
```diff
@@ -20,6 +20,7 @@
 from typing import List, Optional
 
 from ml_space_lambda.data_access_objects.group_user import GroupUserDAO
+from ml_space_lambda.data_access_objects.project_group import ProjectGroupDAO
```
estohlmann, line 23: Remove these?

backend/src/ml_space_lambda/utils/iam_manager.py
```diff
@@ -120,6 +120,23 @@ def __init__(self, iam_client=None, sts_client=None):
...
+            username (str): The username of the user
+
+        Return:
+            Optional[str]: The ARN of the existing role, otherwise None if no such role exists.
+        """
+
+        iam_role_name = self._generate_iam_role_name(username, project_name)
+
+        # Check if the IAM role exists already
+        return self._check_iam_role_exists(iam_role_name)
```
estohlmann, line 138: Not on you but I had to look up what this does:`_check_iam_role_exists`

I was expecting it to return true/false based on the name

backend/src/ml_space_lambda/group/lambda_functions.py
dustins: I think it is right. After deleting the `project_user` association on line 233 it only deletes their dynamic role (user/project combination) if they're not a member of the project through any other means (another group or directly added)

backend/src/ml_space_lambda/project/lambda_functions.py
```diff
@@ -123,18 +129,96 @@ def get(event, context):
...
+
+@api_wrapper
+def add_groups(event, context):
+    project_name = event["pathParameters"]["projectName"]
+    request = json.loads(event["body"])
+    group_names = request["groupNames"]
+    for group_name in group_names:
+        group = group_dao.get(group_name)
+
+        if group:
```
estohlmann, line 173: I wonder if we should check that the ProjectGroupModel doesn't exist for this relationship here too

backend/src/ml_space_lambda/utils/iam_manager.py
dustins: I'm open to renaming it if you have something specific in mind that is a bit more descriptive…

- `_fetch_iam_role`
- `_get_iam_role` (maybe a little confusing with my new `get_iam_role_arn` maybe they're different enough since one is public and the "private"?)

lib/utils/apiFunction.ts
```diff
@@ -67,6 +67,7 @@ export function registerAPIEndpoint (
             DATASETS_TABLE: mlspaceConfig.DATASETS_TABLE_NAME,
             PROJECTS_TABLE: mlspaceConfig.PROJECTS_TABLE_NAME,
             PROJECT_USERS_TABLE: mlspaceConfig.PROJECT_USERS_TABLE_NAME,
+            PROJECT_GROUPS_TABLE: mlspaceConfig.PROJECT_GROUPS_TABLE_NAME,
```
estohlmann, line 70: We need to make sure to add these to the README files and the github docs

backend/src/ml_space_lambda/project/lambda_functions.py
dustins: Good callout. We can skip it if it exists instead of erroring!

frontend/src/entities/project/project.reducer.ts
```diff
@@ -82,9 +83,40 @@ export const listProjectsForUser = createAsyncThunk('project/list_projects_for_u
...
+export const removeGroupFromProject = createAsyncThunk(
+    'user/remove_user_from_project',
+    async (data: IProjectGroup) => {
+        const requestUrl = `/project/${data.project}/groups/${encodeURIComponent(data.group || '')}`;
+        return axios.delete(requestUrl).catch(axiosCatch);
+    }
+);
+
+export const updateProjectGroup = createAsyncThunk(
+    'user/remove_user_from_project',
```
estohlmann, line 113: Copy/Paste fail

frontend/src/entities/group/detail/group-detail.tsx
```diff
@@ -104,17 +108,34 @@ export function GroupDetail () {
                     content: (
                         <Table
                             tableName='Dataset'
-                            actions={groupDatasetUserActions}
+                            actions={GroupDetailDatasetActions}
                             itemNameProperty='name'
                             trackBy='location'
-                            allItems={groupDatasets}
+                            allItems={groupDatasets || []}
```
estohlmann, line 114: Why is this change needed? We aren't doing it for projects or users

backend/src/ml_space_lambda/group/lambda_functions.py
estohlmann: Ah, got ya.

backend/src/ml_space_lambda/utils/iam_manager.py
estohlmann: I think `_fetch_iam_role` works fine. Thanks for fixing!

frontend/src/entities/group/detail/group-detail.tsx
dustins: We don't need that anymore. I think it was leftover from the many revisions I went through on this.

estohlmann (approved):

douglas1850 (approved):

PR #247, opened 2024-07-29
Admin datasets

douglas1850:
*Description of changes:* Admins now have a page where they can view *all* datasets. We've discussed it as a team and agreed that admins should be able to view/edit/delete datasets that they are not the owners of (private) or members of the project for. This will also ensure a dataset (especially Group datasets) can never be orphaned. 


backend/src/ml_space_lambda/data_access_objects/dataset.py
```diff
@@ -139,3 +139,7 @@ def get_all_for_scope(self, dataset_type: DatasetType, scope: str) -> List[Datas
         ).records
 
         return [DatasetModel.from_dict(entry) for entry in json_response]
+
+    def get_all(self) -> List[DatasetModel]:
```
estohlmann, line 143: Probably should write a quick unit test for this

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -317,37 +317,33 @@ def lambda_handler(event, context):
                 if "x-mlspace-dataset-type" in event["headers"] and "x-mlspace-dataset-scope" in event["headers"]:
                     target_type = event["headers"]["x-mlspace-dataset-type"]
                     target_scope = event["headers"]["x-mlspace-dataset-scope"]
-                    if target_type == DatasetType.GLOBAL:
+                    if Permission.ADMIN in user.permissions:
```
estohlmann, line 320: We will want to get Dustin's changes pulled in here so they are accounted for when checking permissions like this

backend/src/ml_space_lambda/dataset/lambda_functions.py
```diff
@@ -305,23 +305,32 @@ def create_dataset(event, context):
 @api_wrapper
 def list_resources(event, context):
     username = event["requestContext"]["authorizer"]["principalId"]
+    is_admin = event["requestContext"].get("resourcePath") == "/admin/datasets"
```
dustins, line 308: This is perfectly valid but it feels a little something to be basing if they're an admin on the URL. Especially with how hard to follow our authorizer is. Not blocking but I'd suggest this be changed to closer to how it is done [in a few other places](https://github.com/awslabs/mlspace/blob/develop/backend/src/ml_space_lambda/group/lambda_functions.py#L67-L69).

```
    user = UserModel.from_dict(json.loads(event["requestContext"]["authorizer"]["user"]))
    if Permission.ADMIN in user.permissions:
```

frontend/src/entities/dataset/detail/dataset-detail.tsx
```diff
@@ -55,7 +62,7 @@ function DatasetDetail () {
     datasetDetails.set('Access level', showAccessLevel(dataset));
     datasetDetails.set('Location', dataset.location);
     if (dataset.type === DatasetType.GROUP) {
-        datasetDetails.set(`Group${dataset.groups && dataset.groups.length > 1 ? 's' : ''}`, dataset.groups);
+        datasetDetails.set(`Group${dataset.groups && dataset.groups.length > 1 ? 's' : ''}`, dataset.groups?.length === 0 ? 'None' : dataset.groups);
```
dustins, line 65: I'm not sure if this actually occurs but it looks like `dataset.groups?.length` could return undefined. In which case it isn't `=== 0` and would try and display `dataset.groups`. Might be better to let truthiness rule the day…

```typescript
dataset.groups?.length ? dataset.groups : 'None'
```

frontend/src/entities/dataset/dataset.tsx
```diff
@@ -65,14 +87,14 @@ export const Dataset = () => {
...
-                    visibleColumns={visibleColumns}
+                    columnDefinitions={isAdmin ? adminDatasetColumns : defaultColumns }
+                    visibleColumns={isAdmin ? visibleAdminColumns : visibleColumns}
                     visibleContentPreference={visibleContentPreference}
                     trackBy='location'
                     itemNameProperty='name'
                     loadingItems={loadingDatasets}
                     loadingAction={loadingAction}
-                    serverFetch={getDatasetsList}
+                    //serverFetch={isAdmin ? getAllDatasets : getDatasetsList}
```
dustins, line 97: Should this be commented out or just not exist?

backend/src/ml_space_lambda/dataset/lambda_functions.py
douglas1850: This is a good point; I think I'll combine the two. I don't want to return all datasets every time an admin tries to list datasets, only when they're listing datasets from the admin datasets page (which uses the /admin/datasets endpoint)

frontend/src/entities/dataset/dataset.tsx
douglas1850: I meant to talk about this in dev sync, I'm trying to get serverFetch to work with `getAllDatasets` or `getDatasetsList` but it's erroring because of typing. Wanted to see if you or Evan knew a way to get this prop to be happy with either function

frontend/src/entities/dataset/detail/dataset-detail.tsx
douglas1850: good point, will do

lib/config.json
```diff
@@ -0,0 +1,7 @@
+{
```
douglas1850, line 1: ah damnit must've committed this on mistake, will remove on next commit

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -426,13 +432,13 @@ def _handle_dataset_request(request_method, path_params, user):
     dataset_name = path_params["datasetName"]
     dataset = dataset_dao.get(dataset_scope, dataset_name)
     if dataset:
-        # Owners can do whatever - this check also handles private datasets
-        if dataset.created_by == user.username:
+        # Owners and Admins can do whatever - this check also handles private datasets
+        if dataset.created_by == user.username or Permission.ADMIN in user.permissions:
```
estohlmann, line 436: nit: this could be simplified to `or  IS_ADMIN`

dustins (approved):

estohlmann (approved):

PR #256, opened 2024-08-05
Remove deprecated constants

douglas1850:
*Description of changes:* The time has come - we've marked several deployment constants as deprecated, and now it's time to actually remove them.

All of these constants will now be handled 100% by the in-app configuration feature.

The only exception is Translation - this can still be managed via app config, but now we also check if the app is being deployed into us-isob-east-1, and if so we disable Translate since it's not available in that region yet.

Did a deployment and nothing failed. Also tried updating my app config with dynamic roles disabled and everything worked. 

Here's how the app config page looks now when dynamic roles are disabled.

Screenshot 2024-08-05 at 5 20 54 PM


dustins (approved):

bin/mlspace-cdk.ts
```diff
@@ -77,12 +77,15 @@ const websiteBucketName = `${config.WEBSITE_BUCKET_NAME}-${config.AWS_ACCOUNT}`;
 const cwlBucketName = `${config.LOGS_BUCKET_NAME}-${config.AWS_ACCOUNT}`;
 const accessLogsBucketName = `${config.ACCESS_LOGS_BUCKET_NAME}-${config.AWS_ACCOUNT}`;
 
+//Translate is not currently available in us-isob-east1
+const enableTranslate = config.AWS_REGION !== 'us-isob-east-1';
```
estohlmann, line 81: We should probably make this an allowlist so that this will work for the future iso regions that might come online without Translate.

frontend/src/entities/configuration/activated-services-configuration.tsx
```diff
@@ -38,6 +38,18 @@ export type ActivatedServicesConfigurationProps = {
     enabledServices: {[key: string]: boolean};
 };
 
+const generateDescription = () => {
+    return `
+        Activate or deactivate services within MLSpace. Deactivated 
+        services will no longer appear within the MLSpace user interface${!window.env.MANAGE_IAM_ROLES ? `.
```
estohlmann, line 44: nit: Since we aren't actually doing anything with the logic here, I would prefer that these ternary operations are positive (remove the !) so it is a bit easier to read.

bin/mlspace-cdk.ts
douglas1850: Okay I'll swap this to a list so we can easily add/remove regions from it as needed

frontend/src/entities/configuration/activated-services-configuration.tsx
douglas1850: Fair point, will swap them around

dustins (dismissed):

estohlmann (approved):

dustins (approved):

PR #260, opened 2024-08-12
Adding support for group membership history

estohlmann:
*Issue #, if available:*

*Description of changes:*

Adding the ability to view group membership history details in the app. This adds entries into the history table any time a user is added/removed or a group is deleted. This will allow admins the ability to do analysis on what users had access to what based on the timeline that is generated.

Screenshot 2024-08-12 at 11 32 42 AM


douglas1850:
>This adds entries into the history table any time a user is added/removed or a group is deleted

Do you have an image for what the history looks like for a group being deleted?

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -302,6 +302,8 @@ def lambda_handler(event, context):
             elif requested_resource == "/group" and request_method == "POST":
                 if IS_ADMIN:
                     policy_statement["Effect"] = "Allow"
+            elif requested_resource == "/group_membership_history" and request_method == "GET":
```
douglas1850, line 305: Is this allowing anyone to get the group membership history?

estohlmann: Yes, it is like the other details items. You can pull it if you have access to the group

douglas1850 (dismissed):

frontend/src/entities/group/detail/group-history.columns.tsx
```diff
@@ -0,0 +1,47 @@
...
+},{
+    id: 'actionedBy',
+    header: 'Actioned By',
+    sortingField: 'actionedBy',
+    cell: (item) => item.actionedBy,
+},{
+    id: 'actionedAt',
+    header: 'Actioned At',
+    sortingField: 'actionedAt',
+    cell: (item) => new Date(item.actionedAt * 1000).toLocaleString(),
```
dustins, line 44: I *think* this should be using https://github.com/awslabs/mlspace/blob/develop/frontend/src/shared/util/date-utils.ts#L167

backend/src/ml_space_lambda/authorizer/lambda_function.py
dustins: Where is it actually checking that they're a member of this group? Also am I reading this wrong or it it checking for `/group_membership_history` when the path of the lambda is `/group-membership-history` (dashes vs hyphens)?

estohlmann: Thank you, I messed it up with my initial deployment, meant to fix it, and forgot 😓

estohlmann: I also updated the authorizer to ensure group membership

frontend/src/entities/group/detail/group-history.columns.tsx
estohlmann: I updated this to use a method in the date-utils class

backend/src/ml_space_lambda/authorizer/lambda_function.py
```diff
@@ -251,9 +251,10 @@ def lambda_handler(event, context):
                                 if project_user:
                                     policy_statement["Effect"] = "Allow"
                 elif "groupName" in path_params:
+                    is_group_member = _is_group_member(path_params["groupName"], username)
                     if IS_ADMIN and request_method in ["POST", "PUT", "DELETE"]:
                         policy_statement["Effect"] = "Allow"
-                    elif request_method == "GET":
+                    elif request_method == "GET" and (IS_ADMIN or is_group_member):
```
dustins, line 257: How would you feel about simplifying this at this point? It looks like if they're an admin, they get a 👍🏻 so then we can remove some of that clutter of checking the method in that scenario. Then we don't have to worry about admins when checking if they're a member.

```python
is_group_member = _is_group_member(path_params["groupName"], username)
if IS_ADMIN:
elif request_method == "GET" and is_group_member:
```

frontend/src/shared/util/date-utils.ts
```diff
@@ -218,23 +218,23 @@ export const hoursToDays = (hours: number | undefined): number | undefined => {
...
  * in the users preferred timezone.
  *
- * @param terminationTime datetime in the form of seconds since epoch
+ * @param epochTime datetime in the form of seconds since epoch
  * @param timeOnly whether or not the display string should be only include the time (notebooks)
  * @returns display friendly string of date time in the users preferred timezone
  */
-export const formatTerminationTimestamp = (terminationTime: number, timeOnly = false) => {
-    const terminationDate = new Date(terminationTime * 1000);
+export const formatEpochTimestamp = (epochTime: number, timeOnly = false) => {
```
dustins, line 228: This is great!

dustins (approved):

douglas1850 (approved):

PR #262, opened 2024-08-13
Num group members

douglas1850:
*Description of changes:* the number of members in a group is now displayed next to a group in the Group table.

Screenshot 2024-08-13 at 5 49 00 PM


backend/src/ml_space_lambda/data_access_objects/group.py
```diff
@@ -35,13 +35,15 @@ def __init__(
         created_by: str,
         created_at: Optional[float] = None,
         last_updated_at: Optional[float] = None,
+        num_members: Optional[int] = None,
```
dustins, line 38: Why are we adding `num_members` do the DAO if it is only computed dynamically in the group listing page?

douglas1850: So that `numMembers` can be populated in the dictionary/JSON we send to the frontend

backend/src/ml_space_lambda/group/lambda_functions.py
```diff
@@ -212,6 +212,9 @@ def list_all(event, context):
     else:
         group_names = [group.group for group in group_user_dao.get_groups_for_user(user.username)]
         groups = group_dao.get_all(group_names=group_names)
+    for group in groups:
+        group_users = group_user_dao.get_users_for_group(group.name)
+        group.num_members = len(group_users)
     return [group.to_dict() for group in groups]
```
dustins, line 218:
```suggestion
    enhanced_groups = [group.to_dict() for group in groups]
    for group in enhanced_groups
        group["num_members"] = len(group_user_dao.get_users_for_group(group.name))
    return enhanced_groups
```

backend/src/ml_space_lambda/data_access_objects/group.py
douglas1850: I like your suggestion, just committed it

douglas1850: I'll make another commit in a minute here removing the numGroup property from the DAO

frontend/src/entities/group/group.columns.ts
```diff
@@ -34,9 +34,15 @@ export const groupColumns: TableProps.ColumnDefinition<IGroup>[] = [{
     header: 'Description',
     sortingField: 'description',
     cell: (item) => item.description,
+},
+{
+    id: 'members',
+    header: 'Members',
+    sortingField: 'members',
+    cell: (item) => item.numMembers || '-',
```
dustins, line 42: I think we have a display function for this. I'll approve now so this can get in the push today but you should follow up on fixing this I think.

estohlmann (approved):

PR #296, opened 2024-12-11
Updated notebook policy to enforce correct system tagging

dustins:
*Issue #, if available:*

*Description of changes:*
We discovered that the Notebook policy was only enforcing the presence of a system tag during resource creation, without validating its value. This PR updates the Notebook policy to require the correct value for the system tag, aligning its enforcement mechanism with the value checks already applied to the user tag in the User policy and the project tag in the Project policy.

### DIFF
--- a/lib/stacks/iam.ts
+++ b/lib/stacks/iam.ts
@@ -73,8 +73,27 @@ export class IAMStack extends Stack {
         // Required tags that force the request to be specific to an MLSpace managed resource
         const requestTagsConditions = {
             'aws:RequestTag/project': 'false',
-            'aws:RequestTag/system': 'false',
+            // this is excluded because it is covered by requestSystemTagEqualsConditions below
+            // 'aws:RequestTag/system': 'false',
             'aws:RequestTag/user': 'false',
+        };
+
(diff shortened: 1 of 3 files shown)


lib/stacks/iam.ts
```diff
@@ -73,9 +73,28 @@ export class IAMStack extends Stack {
         // Required tags that force the request to be specific to an MLSpace managed resource
         const requestTagsConditions = {
             'aws:RequestTag/project': 'false',
-            'aws:RequestTag/system': 'false',
+            // this is excluded because it is covered by requestSystemTagEqualsConditions below
```
estohlmann, line 76: Do we need to have this commented out code here?

estohlmann (approved):

lib/stacks/iam.ts
dustins: I was trying to capture some knowledge so in the future it is obvious to anyone why `system` isn't part of these conditions without an in-depth knowledge of these permissions. If you think that is unnecessary (or worse, confusing) I am happy to remove it.

estohlmann: You can leave it!

PR #315, opened 2025-05-02
Adding Dynamic Configuration Profile management

batzela:
- Add new Lambda handlers (`list_profiles`, `get_profile`, `create_profile`, `update_profile`, `delete_profile`) under `config_profiles/`
  - Use `ApiResponse` and updated `api_wrapper` to support custom status codes and JSON bodies
- Update authorizer to allow GET/POST/PUT/DELETE on `/config-profiles` for admins
- Extend `DynamoDBObjectStore._scan` to accept `projection_expression`
- Add `EnvVariable.CONFIGURATION_PROFILES_TABLE` and update `mlspace_config` defaults
- Define `ResourceInUseError` in exceptions for delete conflicts when a profile is in use
- Update CDK:
  - Deploy `ConfigurationProfilesApiStack` with new REST API endpoints
  - Create `mlspace-ddb-configuration-profiles` table and export its name
  - Add GSI on projects table for `configProfileId` lookups
- Add unit tests:
  - DAO tests for list/get/create/update/delete behavior
  - Lambda handler tests mocking DAO to verify response codes and payloads


backend/src/ml_space_lambda/utils/common_functions.py
```diff
@@ -119,6 +136,12 @@ def wrapper(event, context):
         logger.info(f"Lambda {lambda_func_name}({code_func_name}) invoked with {_sanitize_event(event)}")
         try:
             result = f(event, context)
+            if isinstance(result, ApiResponse):
+                return {
+                    "statusCode": result.status_code,
+                    "headers": result.headers,
+                    "body": result.body
+                }
```
dustins, line 144: 100% up to you but I'd recommend implementing this using inversion of control so the ApiResponse can render itself. It shouldn't matter much with such a limited response object possible for use with APIGW but it'll also clean up the code a little. It would also allow you to have more explicit variants like JsonApiResponse or StringApiResponse instead of hiding special handling based types in the implementation details like is happening on line 65.

```python
            if isinstance(result, ApiResponse):
                 return result.render() // or whatever you want to call it
```

backend/src/ml_space_lambda/utils/exceptions.py
```diff
@@ -24,3 +24,14 @@ def __init__(self, message, http_status_code):
...
+class ResourceInUseError(ServiceException):
+    """
+    Thrown when attempting to delete or modify a resource that is currently in use.
+
+    Attributes:
+        message: error description
+    """
+    def __init__(self, message: str = None):
+        default = "Resource is currently in use and cannot be deleted."
+        super().__init__(message or default, 409)
```
dustins, line 37: Move the default out to a class constant and make it clear in the method signature that is the default.

```suggestion
class ResourceInUseError(ServiceException):
    DEFAULT_MESSAGE = "Resource is currently in use and cannot be deleted."
    """
    Thrown when attempting to delete or modify a resource that is currently in use.

    Attributes:
        message: error description
    """
    def __init__(self, message: str = DEFAULT_MESSAGE):
        super().__init__(message, 409)
```

backend/src/ml_space_lambda/data_access_objects/dynamo_data_store.py
```diff
@@ -108,18 +108,20 @@ def _scan(
...
-            if expression_names:
-                kwargs["ExpressionAttributeNames"] = expression_names
-            if expression_values:
-                kwargs["ExpressionAttributeValues"] = expression_values
+        if projection_expression:
+            kwargs["ProjectionExpression"] = projection_expression
+        if expression_names:
+            kwargs["ExpressionAttributeNames"] = expression_names
+        if expression_values:
+            kwargs["ExpressionAttributeValues"] = expression_values
```
dustins, line 124: Did the comment this removed become invalid/redundant?

backend/src/ml_space_lambda/config_profiles/lambda_functions.py
```diff
@@ -0,0 +1,135 @@
...
+def list_profiles(event, context):
+    """
+    GET /config-profiles
+    Returns list of all profiles (metadata only).
+    """
+    try:
+        items = config_profiles_dao.list()
+        return ApiResponse(
+            body=items,
+            status_code=200
```
dustins, line 26: status_code already defaults to 200 so this is unnecessary

backend/src/ml_space_lambda/config_profiles/lambda_functions.py
```diff
@@ -0,0 +1,135 @@
...
+            training_job_instance_types=payload["trainingJobInstanceTypes"],
+            hpo_job_instance_types=payload["hpoJobInstanceTypes"],
+            transform_job_instance_types=payload["transformJobInstanceTypes"],
+            endpoint_instance_types=payload["endpointInstanceTypes"],
+            profile_id=None,  # DAO will assign
+            created_by=principal_id,
+            created_at=now,
+            updated_by=principal_id,
+            updated_at=now,
+        )
```
dustins, line 78: Could simplify this a lot by doing something like this…

```suggestion
        payload = json.loads(event["body"])
        # Build model and add audit fields
        principal_id = event["requestContext"]["authorizer"]["principalId"]
        payload["created_by"] = principal_id
        payload["updated_by"] = principal_id
        now = int(time.time())
        payload["created_at"] = now
        payload["updated_at"] = now
        profile = ConfigProfileModel(**payload)
```

backend/src/ml_space_lambda/data_access_objects/config_profiles.py
```diff
@@ -0,0 +1,179 @@
...
+
+from ml_space_lambda.data_access_objects.dynamo_data_store import DynamoDBObjectStore
+from ml_space_lambda.enums import EnvVariable
+from ml_space_lambda.utils.exceptions import ResourceInUseError
+from ml_space_lambda.utils.mlspace_config import get_environment_variables
+
+log = logging.getLogger(__name__)
+
+
+class ConfigProfileModel:
```
dustins, line 19: Not something you have to do since it isn't used throughout this project, but we'd really benefit from using pydantic's BaseModel instead of doing all the init and to_dict/from_dict by hand. You can [look at LISA](https://github.com/awslabs/LISA/blob/main/lambda/models/domain_objects.py#L420-L432) to see how simple it can be for the same functionality.

backend/src/ml_space_lambda/config_profiles/lambda_functions.py
```diff
@@ -0,0 +1,135 @@
...
+        result = config_profiles_dao.create(profile)
```
dustins, line 79: Looks like if calling create which sets created_at/updated_at we don't need to set them above either. (although I like the idea of defaulting them to now instead of having to manually set it anywhere)

backend/test/data_access_objects/test_dao_config_profiles.py
```diff
@@ -0,0 +1,154 @@
...
+)
+from ml_space_lambda.data_access_objects.dynamo_data_store import PagedResults
+from ml_space_lambda.utils.exceptions import ResourceInUseError
+
+
+class TestConfigProfilesDAO(unittest.TestCase):
+
+    def setUp(self):
+        # a fake low‑level boto3 client
+        self.mock_client = MagicMock()
```
dustins, line 23: You are 100% following current practices in this project, but it would be great if you had time to implement moto for mocking instead of being forced to mock the AWS SDK.

batzela: Will add this to the backlog.

backend/src/ml_space_lambda/config_profiles/lambda_functions.py
batzela: Updated these so only the DAO handles timestamp modifications.

backend/src/ml_space_lambda/config_profiles/lambda_functions.py
batzela: Love this, incorporated it.

PR #316, opened 2025-05-20
decomposing stacks into constructs

jmharold:
To better support customers deploying into isolated regions, this PR will address a number of issues found while testing isolated deployments:
* Refactor Stacks to use Constructs for creating stack resources.
* Increase memory size of `frontEndDeployment` from `512` to `1024` to resolve out of memory error.
* Remove problematic `us-isob` availability zone from `mlSpaceVPC` causing deployment failures due to unavailable runtimes.
* Setting `restrictDefaultSecurityGroup -> false` in `mlSpaceVPC` to avoid incidental removal of ingress and egress rules attached to security groups.
* Update `generateConfig()` to accept an optional account id and check if any account specific configurations exist before loading the config file generated by `config-helper.mjs`


lib/utils/configTypes.ts
```diff
@@ -241,9 +245,14 @@ export function generateConfig () {
...
-
+    // Try to load account-specific config if an account ID is provided
+    if (accountId && fs.existsSync(`lib/config.${accountId}.json`)) {
+        const accountConfig = JSON.parse(
+            fs.readFileSync(`lib/config.${accountId}.json`).toString('utf8')
+        );
+        _.merge(config, accountConfig);
     //Check for properties set in config.json and default to that value if it exists
-    if (fs.existsSync('lib/config.json')) {
+    } else if (fs.existsSync('lib/config.json')) {
```
dustins, line 255: This looks like both if/else blocks are essentially doing the same thing. Lets either break out that as a function so we're not repeating ourselves or do something like the following.

```typescript
const configPaths = [`lib/config.${accountId}.json`, `lib/config.json`]
for configPath in configPaths {
  if (fx.existsSync(configPath) {
    const fileConfig = JSON.parse(fs.readFileSync(configPath).toString('utf8'));
    _.merge(config, accountConfig);
    break;
  }
}
```

lib/utils/configTypes.ts
```diff
@@ -132,6 +132,9 @@ export type MLSpaceConfig = {
     LAMBDA_ARCHITECTURE: Architecture,
     LAMBDA_RUNTIME: Runtime,
     SYSTEM_ROLE_ARN: string,
+    COMMON_LAYER_PATH?: string,
+    BACKEND_LAMBDA_PATH?: string,
```
dustins, line 136: Is this used anywhere?

lib/stacks/infra/sagemaker.ts
```diff
@@ -31,25 +30,8 @@ export class SagemakerStack extends Stack {
...
-            ],
-            onStart: [
-                {
-                    content: Fn.base64(
-                        readFileSync('lib/resources/sagemaker/lifecycle-start.sh', 'utf8')
-                    ),
-                },
-            ],
-        });
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
dustins, line 33: this seems unnecessary at this point

lib/utils/configTypes.ts
jmharold: Updated in most recent commit to be:

```javascript
    //Try to load account-specific config or config generated by config-helper
    const configPaths = [`lib/config.${accountId}.json`, 'lib/config.json'];
    configPaths.forEach((configPath) => {
        if (fs.existsSync(configPath)) {
            const fileConfig = JSON.parse(fs.readFileSync(configPath).toString('utf8'));
            _.merge(config, fileConfig);
            return;
        }
    });
 ```

